PR #448, opened 2024-02-27
Use heapify instead of sort for heap

rsbowman-striveworks:
Use linear time `heapq.heapify` to create heap.

Chariot PR with the same change: https://github.com/Striveworks/chariot/pull/5779
Slack discussion: https://striveworks.slack.com/archives/C0442QPU63V/p1708989167452269


czaloom (approved):
Thank you @rsbowman-striveworks !

PR #460, opened 2024-02-28
Add Average Recall to OD metrics

ntlind:
## Improvements

Added Average Recall (AR) and Mean Average Recall (mAR) to our OD metrics. To calculate Average Recall (AR):
1. Find the count of true positives above a certain IOU threshold (e.g., .5) for all images containing a ground truth of a particular class.
2. Divide that count of true positives by the total number of ground truths to get the recall per class and IOU threshold. Append that recall value to a list.
3. Repeat steps 1 & 2 for multiple IOU thresholds (e.g., [.5, .75])
4. Take the average of your list of recalls

Note that this metric differs from [COCO's calculation](https://cocodataset.org/#detection-eval) in two ways:
- [COCO calculates](https://github.com/cocodataset/cocoapi/blob/8c9bcc3cf640524c4c20a9c40e89cb6a2f2fa0e9/PythonAPI/pycocotools/cocoeval.py#L466) three different AR metrics (AR@1, AR@5, AR@100) by [limiting the maximum number of detections which are considered during the matching process](https://github.com/rafaelpadilla/review_object_detection_metrics/blob/2efe66d2c4b89e4fcc64e490d4caced2096f03aa/src/evaluators/coco_evaluator.py#L138). Valor doesn't impose any detection limits when creating matched pairs of ground truths and predictions.
- COCO averages across classes while calculating AR, while we calculate AR separately for each class (note this holds true for AP as well: COCO's "Average Precision" is the same calculation as what we call "Mean Average Precision"). Our AR calculations matches the [original FAIR definition](https://arxiv.org/pdf/1502.05082.pdf) of AR, while our mAR calculations match what COCO calls AR.
 
## Tests

Enhanced existing unit, functional, and integration tests. Functional tests match COCO's outputs [here](https://github.com/Lightning-AI/metrics/blob/107dbfd5fb158b7ae6d76281df44bd94c836bfce/tests/unittests/detection/test_map.py#L231).

## Additional References

- [Efficient Graph-Friendly COCO Metric Computation for Train-TimeModel Evaluation](https://arxiv.org/pdf/2207.12120.pdf)
- [What makes for effective detection proposals?](https://arxiv.org/abs/1502.05082)


docs/metrics.md
```diff
@@ -14,16 +13,18 @@ If we're missing an important metric for your particular use case, please [write
...
+| mAP Averaged Over IOUs | The average of several  mAP metrics across class labels. | $\dfrac{1}{\text{number of thresholds}} \sum\limits_{iou \in thresholds} mAP_{iou}$ |
+| Average Recall (AR) | The average of several recall metrics across IOU thresholds, grouped by class labels. | $\dfrac{1}{\text{number of thresholds}} \sum\limits_{iou \in thresholds} \dfrac{TP}{TP + FN}$ |
+| Mean Average Recall (mAR) | The average of several AR metrics across class labels. | $\dfrac{1}{\text{number of classes}} \sum\limits_{class \in classes} AR_{class}$ |
 
-[^1]: When calculating IOUs for object detection metrics, Valor handles the necessary conversion between different types of geometric annotations. For example, if your model prediction is a polygon and your groundtruth is a raster, then the raster will be converted to a polygon prior to calculating the IOU.
+**When calculating IOUs for object detection metrics, Valor handles the necessary conversion between different types of geometric annotations. For example, if your model prediction is a polygon and your groundtruth is a raster, then the raster will be converted to a polygon prior to calculating the IOU.
```
ntlind, line 27: this markdown footnote wasn't rendering correctly, so I changed it to an asterisk

integration_tests/client/filtering/test_geospatial_filtering.py
```diff
@@ -125,7 +125,7 @@ def test_geospatial_filter(
         },
     )
     assert eval_job.wait_for_completion(timeout=30) == EvaluationStatus.DONE
-    assert len(eval_job.metrics) == 9
+    assert len(eval_job.metrics) == 12
```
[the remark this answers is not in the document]
ntlind: we added three new metrics to the output: two AR metrics, and one mAR metric

ekorman:
is there a test that shows we get the same thing using standard tools?

api/valor_api/backend/metrics/detection.py
```diff
@@ -127,7 +133,22 @@ def _ap(
                 )
             )
 
-    return detection_metrics
+            recalls_across_thresholds.append(recalls[-1])
+
+        ar_metrics.append(
+            schemas.ARMetric(
```
ekorman, line 139: where is confidence coming in at computing AR?

ekorman:
can you add to the docs a description of what recall means? confidence has to enter somewhere

api/valor_api/backend/metrics/detection.py
ekorman: this is what the COCO max_dets does: it takes the N most confident predictions and considers those as positive predictions

api/valor_api/schemas/evaluation.py
```diff
@@ -21,6 +21,8 @@ class EvaluationParameters(BaseModel):
         A list of floats describing which Intersection over Union (IoUs) thresholds to calculate a metric for. Must be a subset of `iou_thresholds_to_compute`.
     label_map : LabelMapType, optional
         Optional mapping of individual labels to a grouper label. Useful when you need to evaluate performance using labels that differ across datasets and models.
+    recall_score_threshold: float, optional
```
czaloom, line 24: nit: instead of optional could you write `default=0`

client/valor/coretypes.py
```diff
@@ -1622,7 +1623,8 @@ def evaluate_detection(
             Thresholds to return AP for. Must be subset of `iou_thresholds_to_compute`.
         label_map : Dict[Label, Label], optional
             Optional mapping of individual labels to a grouper label. Useful when you need to evaluate performance using labels that differ across datasets and models.
-
+        recall_score_threshold: float, optional
```
czaloom, line 1626: nit: instead of optional could you write default=0

client/valor/schemas/evaluation.py
```diff
@@ -18,6 +18,8 @@ class EvaluationParameters:
         A list of floats describing which Intersection over Union (IoUs) thresholds to calculate a metric for. Must be a subset of `iou_thresholds_to_compute`.
     label_map: Optional[List[List[List[str]]]]
         Optional mapping of individual labels to a grouper label. Useful when you need to evaluate performance using labels that differ across datasets and models.
+    recall_score_threshold: float, optional
```
czaloom, line 21: nit: instead of optional could you write default=0

docs/metrics.md
```diff
@@ -151,12 +151,12 @@ $\rho_{interp} = \underset{\tilde{r}:\tilde{r} \ge r}{max \ \rho (\tilde{r})}$
...
 3. Repeat steps 1 & 2 for multiple IOU thresholds (e.g., [.5, .75])
 4. Take the average of our list of recalls to arrive at the AR value per class.
 
 Note that this metric differs from COCO's calculation in two ways:
 
 - COCO averages across classes while calculating AR, while we calculate AR separately for each class. Our AR calculations matches the original FAIR definition of AR, while our mAR calculations match what COCO calls AR.
-- COCO calculates three different AR metrics (AR@1, AR@5, AR@100) by limiting the maximum number of detections which are considered during the matching process. Valor doesn't impose any detection limits when creating matched pairs of ground truths and predictions.
+- COCO calculates three different AR metrics (AR@1, AR@5, AR@100) by limiting the maximum number of detections which are considered during the matching process. Valor doesn't impose any detection limits when creating matched pairs of ground truths and predictions. Instead, we allow users to input a `recall_score_threshold` value that will prevent low-confidence predictions from being counted as true positives when calculating AR.
```
ekorman, line 162: nit but I'd prefer wording like

```
COCO calculates three different AR metrics (AR@1, AR@5, AR@100) by considering only the top 1/5/100 most confident predictions during the matching process. Valor, on the other hand, allows users to input a `recall_score_threshold` value that will prevent low-confidence predictions from being counted as true positives when calculating AR.
```

ekorman (approved):

czaloom (approved):

PR #467, opened 2024-02-29
Removed raster dependency on datum

czaloom:
Raster currently requires the datum to encode a matching height and width. This relationship is unintuitive and causes unexpected failures when there is nothing wrong with a users input.


ekorman (approved):

czaloom:
Track raster validation here:

https://github.com/Striveworks/valor/issues/468

PR #469, opened 2024-03-01
PATCH Geometric Conversion

czaloom:


api/tests/functional-tests/backend/core/test_geometry.py
```diff
@@ -27,6 +102,52 @@ def create_clf_dataset(db: Session, dataset_name: str):
     )
 
 
+@pytest.fixture
+def create_objdet_dataset(
```
ntlind, line 106: typo in function name

czaloom: short hand for object detection, ill write it full length

api/tests/functional-tests/backend/core/test_geometry.py
```diff
@@ -92,3 +213,99 @@ def test_convert_geometry(
...
+        )
+    )
+    assert annotation_id is not None
+
+    q = _convert_raster_to_box([])
+    db.execute(q)
+
+    q = _convert_raster_to_polygon([])
+    db.execute(q)
+
```
ntlind, line 250: should we add a test for _convert_raster_to_multipolygon?

czaloom: Raster -> Multipolygon is not implemented as it is a non-trivial conversion. 

See #470 for reference.

ntlind (approved):

PR #478, opened 2024-03-06
add `filter_by` to `Dataset.get_datums`

ekorman:


pyrightconfig.json
czaloom, line 1: Why did you remove these ignores?

czaloom (approved):

PR #479, opened 2024-03-07
Calculate precision-recall curves for classification and object detection tasks

ntlind:
[Link to to PRD](https://docs.google.com/document/d/17Ytay8SR9cF6qRnVdJ-hKj6VoOxLpLUgUQZ2xs9ud5w/edit)


api/valor_api/backend/metrics/classification.py
```diff
@@ -235,38 +372,8 @@ def _compute_confusion_matrix_at_grouper_key(
         that have both a ground truth and prediction with label key `label_key`. Otherwise
         returns the confusion matrix.
     """
-    label_key_filter = list(
-        grouper_mappings["grouper_key_to_label_keys_mapping"][grouper_key]
-    )
 
-    # groundtruths filter
-    gFilter = groundtruth_filter.model_copy()
```
ntlind, line 243: I moved this section out of this function and into the preceding scope so that I could use these objects for `_compute_curves`.

api/valor_api/backend/metrics/detection.py
```diff
@@ -312,9 +415,9 @@ def _annotation_type_to_column(
             pd.c.annotation_id.label("pd_ann_id"),
             pd.c.score.label("score"),
         )
-        .select_from(gt)  # type: ignore - SQLAlchemy type issue
-        .join(
-            pd,  # type: ignore - SQLAlchemy type issue
+        .select_from(pd)  # type: ignore - SQLAlchemy type issue
+        .outerjoin(
```
ntlind, line 419: we make this an `outerjoin` so we can identify false predictions that aren't attached to any ground truths via the `false_positive_entries` object. this operation doesn't affect the existing AP/AR metrics since we filter out these cases from those calculations in line 465.

integration_tests/client/metrics/test_classification.py
```diff
@@ -810,8 +810,660 @@ def test_evaluate_classification_with_label_maps(
...
 
     cat_expected_metrics = [
+        {"type": "Accuracy", "parameters": {"label_key": "k3"}, "value": 0.0},
+        {"type": "ROCAUC", "parameters": {"label_key": "k3"}, "value": 1.0},
+        {
+            "type": "PrecisionRecallCurve",
+            "parameters": {"label_key": "k3", "pr_curve_iou_threshold": None},
+            "value": {
+                "v1": {
+                    "0.1": {
```
ntlind, line 820: the confidence threshold keys are stored as strings instead of floats in psql. this isn't ideal since it causes these keys to be out-of-order when viewed via the Python client, but I didn't see any easy workarounds to this issue given we're using the JSONB field type. I'd recommend that we leave as-is and let the front-end team convert these keys to floats on their end, but open to your feedback

integration_tests/client/metrics/test_detection.py
```diff
@@ -943,31 +955,31 @@ def test_evaluate_detection_with_label_maps(
...
+            "value": -1.0,
+            "label": {"key": "class", "value": "cat"},
         },
         {
             "type": "AR",
             "parameters": {"ious": [0.1, 0.6]},
-            "value": -1.0,
-            "label": {"key": "class", "value": "cat"},
+            "value": 0.0,
+            "label": {"key": "class", "value": "siamese cat"},
```
ntlind, line 982: note that the order of some of these expected_metrics has changed, but the values themselves are the same as what they were before.

ekorman (approved):

migrations/sql/00000001_initialize_schema.up.sql
czaloom, line 1: I believe we are implementing migrations as of the open source release. Could you move this chnage into its own migration file.

api/valor_api/backend/metrics/detection.py
czaloom, line 1: Im a little confused, shouldn't the PR curve for object detection be per `label` not per `label_key` ?

PR #486, opened 2024-03-15
Implement quick follow-ups on precision-recall curve PR

ntlind:
## Improvements
- Remove unnecessary argument (`grouper_ids_associated_with_gts`) from `_calculate_ap_ar`
- Add docs for `PrecisionRecallCurve`
- Create separate migration files for converting `value` in the `metric` table to a JSONB


czaloom (approved):

migrations/sql/00000002_alter_metric_value_type.down.sql
czaloom, line 1: Why is there no need to revert?

ekorman (approved):

PR #488, opened 2024-03-17
add typescript client

ekorman:


ekorman:
ci/cd publishes to https://www.npmjs.com/package/@striveworks/valor upon making a release (similar to the python client)

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,163 @@
...
+  | 'empty'
+  | 'classification'
+  | 'object-detection'
+  | 'semantic-segmentation'
+  | 'embedding';
+
+type Evaluation = {
+  id: number;
+  model_name: string;
+  datum_filter: Dataset;
```
czaloom, line 26: Why is this a Dataset type?

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,163 @@
...
+    this.client = axios.create({
+      baseURL,
+      headers: {
+        'Content-Type': 'application/json'
+      }
+    });
+  }
+
+  public async getDatasets(queryParams: object): Promise<Dataset[]> {
+    const response = await this.client.get('/datasets', { params: queryParams });
```
ntlind, line 57: nit: is it worth adding docstrings to these methods (similar to the ones we use in our Python client)?

ts-client/src/ValorClient.ts
ekorman: good catch, added

ts-client/src/ValorClient.ts
ekorman: yea i think so-- i just added

.github/workflows/build-and-publish.yml
```diff
@@ -23,3 +23,21 @@ jobs:
...
+  build-and-publish-ts-package:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./ts-client
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "16"
```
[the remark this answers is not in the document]
ekorman: yea github was warning that its old too, just bumped it

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,334 @@
+import axios, { AxiosInstance } from 'axios';
+
+type Dataset = {
+  id: number;
+  name: string;
+  metadata: object;
```
weirong-duke, line 6: metadata is an undefiend object of some structure?

i dont think you're using metadata anywhere in Valor but `Partial<Record<string, any>>` might be a safer type here just in case future users do

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,334 @@
...
+   * @param datasetName name of the dataset
+   * @param datumUid uid of the datum
+   * @param annotations annotations to add
+   *
+   * @returns {Promise<void>}
+   */
+  public async addGroundTruth(
+    datasetName: string,
+    datumUid: string,
+    annotations: object[]
```
weirong-duke, line 303: we dont have more specific typing we can use here?

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,334 @@
...
+  | 'semantic-segmentation'
+  | 'embedding';
+
+type Evaluation = {
+  id: number;
+  model_name: string;
+  datum_filter: object;
+  parameters: { task_type: TaskType; object };
+  status: 'pending' | 'running' | 'done' | 'failed' | 'deleting';
+  metrics: object[];
```
weirong-duke, line 29: are these not structured?

weirong-duke (commented):
okay all `object` typing related questions

ts-client/src/ValorClient.ts
ekorman: metadata is very generic, i'll switch to that type

ts-client/src/ValorClient.ts
ekorman: they are but its just a tone of types that it requires. i originally had these https://github.com/Striveworks/valor/blob/659952aa9baf7cb14a5d2da1cb8ae283f0e17eb2/ts-client/src/ValorClient.ts but got rid of them to simplify since this method is only used for testing purposes-- i don't think there will be any case (at least soon) where the UI will be creating groundtruth. but happy to add them back if you think that makes sense

ts-client/src/ValorClient.ts
jyono: `Record<string, any>` might be good here because we always expect metadata to come back, it may or may not be an empty object?

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,337 @@
...
+   * because we define higher-level methods that use this.
+   *
+   * @param queryParams An object containing query parameters to filter datasets by.
+   *
+   * @returns {Promise<Dataset[]>}
+   *
+   */
+  private async getDatasets(queryParams: object): Promise<Dataset[]> {
+    const response = await this.client.get('/datasets', { params: queryParams });
+    return response.data;
```
jyono, line 84:
```suggestion
    return this.client.get('/datasets', { params: queryParams }).data;
```
await is implied when returning from an async function. if the call fails, the promise will reject basically the same as it would've.

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  /**
+   * Fetches a model given its name
+   *
+   * @param name name of the model
+   *
+   * @returns {Promise<Model>}
+   */
+  public async getModelByName(name: string): Promise<Model> {
```
jyono, line 206: missing a test(if you care)

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,337 @@
...
+    annotations: object[]
+  ): Promise<void> {
+    await this.client.post('/predictions', [
+      {
+        model_name: modelName,
+        datum: { uid: datumUid, dataset_name: datasetName },
+        annotations: annotations
+      }
+    ]);
+  }
```
jyono, line 349: better to always return the response IMO so clients have the option to do something with it.

ts-client/src/ValorClient.ts
ekorman: they are a bit more structured-- just added more typing

.github/workflows/build-and-publish.yml
```diff
@@ -23,3 +23,21 @@ jobs:
...
+        with:
+          node-version: "20"
+          registry-url: "https://registry.npmjs.org"
+      - name: Set version from the git tag
+        run: npm version ${{ github.ref_name }}
+      - run: npm ci
+      - name: Publish to npm
+        run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_API_TOKEN }}
```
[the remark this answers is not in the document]
ekorman: i was following this https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages which used `NODE_AUTH_TOKEN` and i tested that this workflow runs succesfully

ts-client/src/ValorClient.ts
ekorman: updated in [9b1912a](https://github.com/Striveworks/valor/pull/488/commits/9b1912a2347783be4eccbd91571bc53adac1233c)

ts-client/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/striveworks/valor.git"
+  },
+  "keywords": [
+    "valor",
+    "striveworks"
+  ],
+  "author": "Striveworks <valor@striveworks.com>",
```
jyono, line 18: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#people-fields-author-contributors

add ref to striveworks.com? 

```suggestion
  "author": "Striveworks <valor@striveworks.com> (https://www.striveworks.com/)",
```

jyono (approved):
lgtm assuming others are resolved

TaiHovanky:
Does there need to be a /dist folder containing the built package? I tried installing it into Horus, but I get this error ` Cannot find module '@striveworks/valor' or its corresponding type declarations.` even though the @striveworks/valor folder shows up in my node_modules.

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,350 @@
+import axios, { AxiosInstance } from 'axios';
+
+type Dataset = {
```
TaiHovanky, line 3: These types might need to be exported so that I can import them into Horus.

ts-client/src/ValorClient.ts
```diff
@@ -0,0 +1,350 @@
...
+  id: number;
+  name: string;
+  metadata: Partial<Record<string, any>>;
+};
+
+type Model = {
+  id: number;
+  name: string;
+  metadata: object;
```
TaiHovanky, line 12: metadata's type should be changed to `Partial<Record<string, any>>`

ts-client/src/ValorClient.ts
ekorman: done

ts-client/src/ValorClient.ts
ekorman: updated!

TaiHovanky (approved):

PR #489, opened 2024-03-18
Improve test coverage to 90% and enforce coverage check in `combine-coverage-report` workflow

ntlind:
## Improvements
- Add more tests to get coverage back to 90%
- Force `combine-coverage-report` workflow to error out if test coverage is below 90%

## Testing
- Tested the new `combine-coverage-report` check in [this commit](https://github.com/Striveworks/valor/actions/runs/8349311881/job/22853317813)


client/valor/client.py
```diff
@@ -122,6 +122,13 @@ def chunk_list(
     n_elements = len(json_list)
     avg_element_size = get_json_size(json_list) / n_elements
     n_elements_per_chunk = int(chunk_size_bytes / avg_element_size)
+
+    # handle division by zero error:
+    if n_elements_per_chunk == 0:
```
czaloom, line 127: Could you instead check the input args as a divide by zero is still possible.

```
if (n_elements > 0) and (chunk_size_bytes > 0):
        raise ValueError(
            "Divison by zero error. Please increase your chunk_size_bytes parameter."
        )
```

Given input validation, instead of checking `n_elements_per_chunk` we can instead write 

```
n_elements_per_chunk = int(chunk_size_bytes // avg_element_size) + 1
```

ntlind: good catch, thank you! I didn't follow why we need `n_elements_per_chunk = int(chunk_size_bytes // avg_element_size) + 1`; take a look at the recent update and see if you agree with it

czaloom: I realized another problem with raising an error is that having an object exceed the chunk size should be handled by uploading it individually. raising an error will prevent this behavior.

wrt to the new test, with an undersized `chunk_size_bytes` param, the desired behavior here would NOT be raise an error. It would be to return the list chunked into a list of single element lists.
```
# 100 elements with an average element size of 23.8
    data = [{f"key_{i}": f"value_{i}"} for i in range(100)]

    # division by zero error
        _chunk_list(json_list=data, chunk_size_bytes=1)
```

czaloom (approved):

Issue #491, opened 2024-03-19
BUG: race condition when deleting models and datasets


ekorman:
### valor version checks

- [X] I have confirmed this bug exists on the latest version of valor.

### Reproducible Example

there can be a race condition when putting in requests to delete a model and dataset when there's an evaluation on that model via the dataset

this happens sometimes in the typescript test teardown:
https://github.com/Striveworks/valor/blob/487-enh-typescript-client/ts-client/tests/ValorClient.test.ts#L22

```javascript
afterEach(async () => {
  // delete any datasets or models in the backend
  const datasets = await client.getAllDatasets();
  await Promise.all(
    datasets.map(async (dataset) => {
      await client.deleteDataset(dataset.name);
    })
  );

  const models = await client.getAllModels();
  await Promise.all(
    models.map(async (model) => {
      await client.deleteModel(model.name);
    })
  );

  // wait for all models and datasets to be deleted
  while (
    (await client.getAllModels()).length > 0 &&
    (await client.getAllDatasets()).length > 0
  ) {
    await new Promise((resolve) => setTimeout(resolve, 100));
  }
});
```

### Issue Description

backend error message attached
[err.txt](https://github.com/Striveworks/valor/files/14654318/err.txt)

### Expected Behavior

the backend should be able to handle this

ekorman:
this asynchronous code also caused a 500 error
```javascript
await Promise.all(
    datasetNames.map(async (datasetName) => {
      await client.createDataset(datasetName, {});
      await client.addGroundTruth(datasetName, 'uid1', [
        {
          task_type: 'classification',
          labels: [{ key: 'label-key', value: 'label-value' }]
        }
      ]);
      await client.finalizeDataset(datasetName);
    })
  );
 ```

PR #495, opened 2024-03-21
evaluation deletion race condition

ekorman:
fixes bug in https://github.com/Striveworks/valor/issues/491


ekorman:
another way i tested this:

one way i consistently recreated the bug was changing the backend code so that when `delete_evaluations` was called with datasets, it would sleep in between the `fetch_evaluations` call and iterating through those to delete. and then i would run this script:

```
def delete_dataset(dset_name):
    try:
        client.delete_dataset(dset_name)
        print(f"Deleted dataset: {dset_name}")
    except Exception as e:
        print(f"Failed to delete dataset {dset_name}: {e}")


def delete_model(model_name):
    try:
        client.delete_model(model_name)
        print(f"Deleted model: {model_name}")
    except Exception as e:
        print(f"Failed to delete model {model_name}: {e}")


if __name__ == "__main__":
    for dset_name in dset_names:
        delete_dataset(dset_name)
    for model_name in model_names:
        delete_model(model_name)

    create()
    with ThreadPoolExecutor(4) as p:
        p.map(delete_dataset, dset_names)
        # time.sleep(5)
        p.map(delete_model, model_names)
```

which would give the race condition of the same evaluation trying to be deleted by the dataset and model it corresponds to. this script now passes (when now putting a sleep between `fetch_evaluations_and_mark_for_deletion` and the deletion calls).

api/valor_api/backend/core/evaluation.py
```diff
@@ -489,7 +492,21 @@ def fetch_evaluations(
...
+            )
+        )
+        .values(status=enums.EvaluationStatus.DELETING)
+        .execution_options(synchronize_session="fetch")
+    )
+
+    return db.execute(stmt).scalars().all()
 
 
 def fetch_evaluation_from_id(
```
czaloom, line 512: We should add a condition to this query so that it doesn't return evaluations that are marked for deletion

czaloom: This would go for all of the get_evaluation methods too

ekorman: i'm not sure i understand-- we want this to return all the ones that were newly marked for deletion

api/valor_api/backend/core/evaluation.py
```diff
@@ -459,14 +461,15 @@ def create_or_get_evaluations(
     )
 
 
-def fetch_evaluations(
+def fetch_evaluations_and_mark_for_deletion(
```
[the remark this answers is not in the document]
ekorman: yea i agree, ill do that

czaloom (approved):

PR #502, opened 2024-03-27
Update `PrecisionRecallCurve` metric to allow for datum visualization

ntlind:
## Improvement
- Update the `PrecisionRecallCurve` output to contain all of the details needed to fetch and visualize datums from the Dataset service. The new output looks like:

```python
{
    "type": "PrecisionRecallCurve",
    "parameters": {
        "label_key": "class", # The key of the label.
        "pr_curve_iou_threshold": 0.5, # Note that this value will be None for classification tasks. For detection tasks, we use 0.5 as the default threshold, but allow users to pass an optional `pr_curve_iou_threshold` parameter in their evaluation call.
    },
    "value": {
        "cat": { # The value of the label.
            "0.05": { # The confidence score threshold, ranging from 0.05 to 0.95 in increments of 0.05.
                "fn": [],
                "fp": [
                    (
                        'test_dataset',
                        1,
                        '{"type":"Polygon","coordinates":[[[464.08,105.09],[495.74,105.09],[495.74,146.99],[464.08,146.99],[464.08,105.09]]]}'
                    ) # There's one false positive for this (key, value, confidence threshold) combination as indicated by the one tuple shown here. This tuple contains that observation's dataset name, datum ID, and coordinates in the form of a GeoJSON string. For classification tasks, this tuple will only contain the given observation's dataset name and datum ID.
                ],
                "tp": [],
                "recall": -1,
                "f1_score": -1,
                "precision": 0.0,
            },
            ...
```


api/valor_api/backend/metrics/detection.py
```diff
@@ -98,37 +103,56 @@ def _compute_curves(
         for confidence_threshold in [x / 100 for x in range(5, 100, 5)]:
             if grouper_id not in sorted_ranked_pairs:
                 # happens when there are no ground truths for a label
-                tp = 0
+                tp = []
                 fn = (
```
czaloom, line 107: Shouldn't this be a false-positive if there is no groundtruth for the label?

czaloom: Is grouper_id indicative of groundtruths or predictions?

ntlind: `sorted_ranked_pairs` contains ground truth + prediction matches. if there's no grouper_id in that object, then there weren't any qualifying ground truth + prediction matches for that grouper label (and hence no true positives)

ntlind: I agree that comment is confusing, I'll remove it

api/valor_api/backend/metrics/detection.py
```diff
@@ -98,37 +103,56 @@ def _compute_curves(
...
+                            (row.dataset_name, row.gt_datum_id, row.gt_geojson)
+                        ]
+                        seen_gts.add(row.gt_id)
+
+                fn = [
+                    (dataset_name, datum_id, gt_geojson)
+                    for dataset_name, datum_id, gt_id, gt_geojson in groundtruths_per_grouper[
+                        grouper_id
+                    ]
+                    if gt_id not in seen_gts
```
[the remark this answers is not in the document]
ntlind: yep, lack of overlap with a qualified prediction (one with an IOU > threshold and score > threshold)

api/valor_api/backend/metrics/detection.py
```diff
@@ -378,6 +405,11 @@ def _annotation_type_to_column(
                 grouper_mappings["label_id_to_grouper_id_mapping"],
                 value=models.GroundTruth.label_id,
             ).label("label_id_grouper"),
+            (
+                gfunc.ST_AsGeoJSON(
```
czaloom, line 408: `ST_AsGeoJSON` does not support the raster type. 

There are two options here...
1. Returning the original annotation will require using the same encoding we to load schemas.
2. Cast it to a envelope and only ever return bounding boxes.

api/valor_api/backend/metrics/detection.py
```diff
@@ -395,15 +428,25 @@ def _annotation_type_to_column(
                 grouper_mappings["label_id_to_grouper_id_mapping"],
                 value=models.Prediction.label_id,
             ).label("label_id_grouper"),
+            (
+                gfunc.ST_AsGeoJSON(
```
czaloom, line 431: Same issue as above. `ST_AsGeoJSON` does not support raster.

api/valor_api/backend/metrics/detection.py
```diff
@@ -517,21 +576,36 @@ def _annotation_type_to_column(
                 grouper_mappings["label_id_to_grouper_id_mapping"],
                 value=models.GroundTruth.label_id,
             ).label("label_id_grouper"),
+            models.Datum.id.label("datum_id"),
+            models.Dataset.name.label("dataset_name"),
+            (
+                gfunc.ST_AsGeoJSON(
```
czaloom, line 581: Same issue as above. ST_AsGeoJSON does not support raster.

api/tests/functional-tests/backend/metrics/test_detection.py
czaloom, line 1: Where are these hashes computed? Im a little concerned that this will lead to issues later on as python's `hash` is not deterministic over time.

api/valor_api/schemas/metrics.py
```diff
@@ -411,13 +411,26 @@ class PrecisionRecallCurve(BaseModel):
...
+            dict[
+                str,
+                int
+                | float
+                | list[tuple[str, int]]  # for classification tasks
+                | list[tuple[str, int, str]]  # for object detection tasks
+                | None,
+            ],
+        ],
+    ]
```
czaloom, line 433: nit: Im a little wary of how complicated this is getting without any description of what each layer is in a value.

api/tests/functional-tests/backend/metrics/test_detection.py
czaloom: https://stackoverflow.com/questions/27954892/deterministic-hashing-in-python-3

api/valor_api/backend/metrics/detection.py
ntlind: thanks for this callout! I used `ST_Envelope` on rasters in `detection.py`, and added tests to evaluate whether object detection works as expected for annotations with rasters.

api/tests/functional-tests/backend/metrics/test_detection.py
ntlind: they are pulled directly from one local run of test__compute_detection_metrics

the hashes themselves change with each run, but that doesn't matter for this test as they're being hardcoded in the input arguments

api/valor_api/schemas/metrics.py
ntlind: what change would you recommend given we want to offer GeoJSON strings for OD tasks, but not for classification tasks?

czaloom (approved):

PR #505, opened 2024-03-28
Return null when metric value is unknown.

czaloom:


api/valor_api/schemas/metrics.py
```diff
@@ -25,6 +25,27 @@ class Metric(BaseModel):
     value: float | dict | None = None
     label: Label | None = None
 
+    @field_serializer("value")
+    def serialize_negative_metric_value(value: float | dict | None):  # type: ignore - not sure why
```
ntlind, line 29: should we update the outputs in `detection.py`, `classification.py`, etc. to return `None` instead of `-1` for all relevant values, rather than converting the -1's to `None` later in this process?

PR #507, opened 2024-04-01
Better logs for evaluation.

czaloom:


api/valor_api/backend/metrics/detection.py
```diff
@@ -868,11 +868,19 @@ def compute_detection_metrics(*, db: Session, evaluation_id: int):
         .all()
     )
     model = (
-        db.query(Query(models.Model).filter(prediction_filter).any())  # type: ignore - SQLAlchemy type issue
```
ntlind, line 871: is `prediction_filter` guaranteed to only include this one check (`models.Model.name == evaluation.model_name`)?

api/valor_api/backend/metrics/detection.py
```diff
@@ -868,11 +868,19 @@ def compute_detection_metrics(*, db: Session, evaluation_id: int):
...
-        .distinct()
+        db.query(models.Model)
+        .where(models.Model.name == evaluation.model_name)
         .one_or_none()
     )
 
+    # verify
+    if not datasets:
+        raise RuntimeError("No datasets returned by datum filter.")
```
ntlind, line 878: could you add a test to check that this branching logic works as expected?

api/valor_api/backend/metrics/metric_utils.py
```diff
@@ -294,8 +294,10 @@ def wrapper(*args, **kwargs):
             core.set_evaluation_status(
                 db, evaluation_id, enums.EvaluationStatus.FAILED
             )
-            logger.info(
-                f"Evaluation `{evaluation_id}` failed with error: {str(e)}"
+            logger.error(
+                "Valor Evaluation Exception",
```
ntlind, line 298: should we include the `evaluation_id` in the error message?

api/valor_api/backend/metrics/detection.py
czaloom: Reverting this

ntlind (approved):

PR #510, opened 2024-04-02
Fix `make dev-env` for Mac M1/M2 users

ntlind:
## Issue

Running `make dev-env` fails for Mac M1/M2 users since we switched from `-slim` to `-alpine` in #490. There are a few build errors when running this command, but the most glaring one involves `scipy` (which is a requirement for `motmetrics`). These build failures do not affect production, but may confuse new users who want to try out Valor on their Macbooks.

## Recommendation
- Delete `multi_object_tracking.py` and its associated unit-tests so that we can remove `motmetrics` (which depends on `scipy`) from our dependency lists. Removing this functionality seems like the right move given a) we don't advertise these capabilities / metrics in our docs or sample notebooks and b) these functions aren't well-tested, and likely can't be used out-of-the-box in their current state.
- Install `gcc` and `libffi-dev` to resolve remaining `alpine` dependency issues

## Alternatives
- **Switch back from `-alpine` to `-slim`**: `-slim` definitely seems to be the more popular image choice for services like ours, and would solve this issue without removing `motmetrics`. [John mentioned](https://striveworks.slack.com/archives/C04JJ15357U/p1712081705184079?thread_ts=1711986706.116799&cid=C04JJ15357U) that `-alpine` is patched much more often than `-slim`, so there's some security risk associated with this option.
- **Throw a warning for M1/M2 users to use `make start-postgres-docker && make run-migrations` when they try to call `make dev-env`**:  This option is a bit of a cop out, but is the easiest to implement while still resolving the core issue above.
- **Continue investigating how to satisfactory install `scipy` to resolve `motmetrics` dependency issue**: The `scipy` dependency issue is theoretically solvable, but isn't worth the time IMO. John and I tested 10-15 different fixes today, but didn't find one that worked.


ntlind:
> Also, is removing scipy just delaying the problem since we could potentially need it in the future?

I don't think we will, but kicking this problem down the road until we do seems wise.

ekorman (approved):

PR #514, opened 2024-04-04
Added symbolic version of `typing.Optional`

czaloom:
### Feature Description

Implement a symbolic equivalent of `typing.Optional` in the same manner `valor.schemas.List` was implemented to replace `typing.List`.

The new type will be called `valor.schemas.Nullable`

This will allow the removal of `BoundingBox` and `BoundingPolygon` which are represented by `Optional[Polygon]` under the hood.

Typical usage will look like the following...
```
# static collections automate type assignment
class Annotation(StaticCollection):
    task_type: TaskTypeEnum
    labels: List[Label]
    metadata: Dictionary
    box: Nullable[Box]
    polygon: Nullable[Polygon]
    raster: Nullable[Raster]
    embedding: Nullable[Embedding]

# inline usage
nullable_float = Nullable[Float](0.5)
nullable_float = Nullable[Float](None)
```

### DIFF
--- a/client/valor/schemas/compatibility.py
+++ b/client/valor/schemas/compatibility.py
@@ -1,14 +1,10 @@
 from typing import Any, Optional, Union
 
-from valor.schemas.symbolic.collections import (
-    Dictionary,
-    _get_atomic_type_by_name,
-)
 from valor.schemas.symbolic.types import (
-    BoundingBox,
-    BoundingPolygon,
+    Box,
     Date,
     DateTime,
+    Dictionary,
     Duration,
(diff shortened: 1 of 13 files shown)


client/valor/schemas/compatibility.py
```diff
@@ -18,6 +14,7 @@
     Polygon,
     Raster,
     Time,
+    _get_type_by_name,
```
ntlind, line 17: change this name to indicate this is a public method

ntlind (approved):

PR #516, opened 2024-04-05
BugFix: PrecisionRecallCurve Datum UID

czaloom:
### Issue Description

The `datum_id` provided by the `PrecisionRecallCurve` metric is the sql row id and not the datum UID. As this is an internal identifier the client is unable to use this value.

### Expected Behavior

The `datum_id` field in the returned metrics should be the datum UID.


api/valor_api/backend/metrics/classification.py
```diff
@@ -102,37 +122,43 @@ def _compute_curves(
...
+                groundtruths_with_uids.c.datum_id,
+                groundtruths_with_uids.c.datum_uid,
+                groundtruths_with_uids.c.dataset_name,
             )
         )
 
+        print("==================== TEST 1")
         res = list(db.execute(total_query).all())
-
+        print("==================== TEST 2")
```
ntlind, line 155: remove print statements

api/tests/functional-tests/backend/metrics/test_detection.py
```diff
@@ -160,6 +182,8 @@ def test__compute_curves(db: Session):
                 dataset_name="test_dataset",
                 pd_datum_id=527,
```
ntlind, line 183: should we remove `datum_id`s since they can't be used externally?

api/valor_api/backend/metrics/classification.py
```diff
@@ -102,37 +122,43 @@ def _compute_curves(
...
         # handle edge case where there were multiple prediction labels for a single datum
         # first we sort, then we only increment fn below if the datum_id wasn't counted as a tp or fp
         res.sort(
             key=lambda x: ((x[1] is None, x[0][0] != x[0][1], x[1], x[2]))
         )
+        print("==================== TEST 3")
```
ntlind: remove

api/valor_api/backend/metrics/detection.py
```diff
@@ -25,6 +25,8 @@ class RankedPair:
     dataset_name: str
     pd_datum_id: int
     gt_datum_id: int
```
ntlind, line 27: should we remove these?

api/tests/functional-tests/backend/metrics/test_detection.py
czaloom: RankedPairs require the actual id b/c they are also used in calculating AP. They are internal storage and dont reflect what is returned to the user.

The datum_id position in the returned metric is now populated by the datum UID instead.

czaloom: The returned tuple is (dataset_name, datum_uid, geojson)

ntlind: aw that's right, thanks for clarifying

api/valor_api/backend/metrics/classification.py
```diff
@@ -82,10 +83,29 @@ def _compute_curves(
...
             )
+            .join(
+                models.Datum,
+                models.Datum.id == models.Annotation.datum_id,
+            )
             .where(predictions.c.score >= threshold)
             .alias()
         )
 
+        groundtruths_with_uids = (
```
ntlind, line 94: could we add the changes here to the original `groundtruths` instead of creating a new subquery?

ntlind (approved):

PR #517, opened 2024-04-09
Add parameters to TS client

ntlind:
## Improvements

- Fix API schema to allow `recall_score_threshold` and `pr_curve_iou_threshold` to be optional
- Add optional parameters to `createOrGetEvaluation` and `bulkCreateOrGetEvaluations` to bring the TS client to parity with the python client. This change is needed for the EvaluationUX initiative to calculate precision-recall curves.

Example:
```js
public async createOrGetEvaluation(
    model: string,
    dataset: string,
    taskType: TaskType,
    // new parameters
    iouThresholdsToCompute?: number[],
    iouThresholdsToReturn?: number[],
    labelMap?: number[][][],
    recallScoreThreshold?: number,
    computePrCurves?: boolean,
    prCurveIouThreshold?: number
  ): ...
```

## Testing

Added one test to check that `PrecisionRecallCurve` is returned correctly when the boolean is set to True.

PR #519, opened 2024-04-09
Fix TS tests by lengthening timeouts

ntlind:
## Improvements

Increase timeout durations to fix failing tests after we merged #517


ts-client/tests/ValorClient.test.ts
czaloom, line 1: Is the timeout in ms or seconds?

czaloom (approved):

PR #521, opened 2024-04-09
Removed "no annotations" warning.

czaloom:


integration_tests/client/datasets/test_groundtruth.py
```diff
@@ -197,20 +197,18 @@ def test_add_groundtruth(
...
-                datum=Datum(
-                    uid="uid",
-                    metadata={
-                        "height": 200,
-                        "width": 150,
-                    },
-                ),
-                annotations=[],
-            )
+    dataset.add_groundtruth(
```
ekorman, line 200: whats the point of this (and similary the empty `add_prediction` below) now? is it to verify nothing goes wrong when adding an empty groundtruth or prediction?

ekorman: if we want to check that no warnings are thrown we can do

```
with warnings.catch_warnings():
    warnings.simplefilter("error")
    dset.add_groundtruth(...)
```

(see second bullet point here https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests)

ekorman (approved):

PR #530, opened 2024-04-11
Add pagination for `GET` endpoints

### LINKED ISSUE #512: ENH: Pagination for /evaluations endpoint and others
### Feature Type

- [ ] Adding new functionality to valor

- [X] Changing existing functionality in valor

- [ ] Removing existing functionality in valor

### Problem Description

There may be many evaluations corresponding to a model or dataset, too many to fit in a single response.  It would be nice to have pagination support for GET `\evaluations`.  Same goes for `\models`, `\datasets`, and possibly other endpoints.

### Feature Description

These endpoints could take new `start` and `limit` params.   At most `limit` results would be returned per request, and the items would begin at offset `start` from the entire result set.

### Additional Context

_No response_

ntlind:
## Improvements

Adds pagination to most of our `GET` endpoints:
- Users pass in `offset` (the index number of the first item to return) and `limit` (the number of items to return) as query params (e.g., `/models?offset=0&limit=25)`
- The items are then sorted by `created_at` (newer items towards the top) and limited in SQL
- The items are then returned with a `content-range` header indicating the number of items returned and the total number of returnable items (e.g., `content-range: items 0-24/66)`). If there aren't any items to return, the header will be `content-range: items */0)`.

Affected endpoints:
- `/models`
- `/datasets`
- `/evaluations`
- `/datums`
- `/labels`

Resolves #512.

## Testing

Implemented new functional and integration tests to check new response.


api/valor_api/backend/core/label.py
```diff
@@ -320,6 +320,80 @@ def get_labels(
     }
 
 
+def get_paginated_labels(
```
ntlind, line 323: I made this a new function since `get_labels` is used both for the end point and for a bunch of internal calls. The headers are only useful for the endpoints, so it makes sense to split `get_labels` from `get_paginated_labels`.

api/tests/functional-tests/backend/core/test_evaluation.py
```diff
@@ -421,13 +421,47 @@ def test_get_evaluations(
...
+
+    evaluations, headers = core.get_evaluations(
+        db=db,
+        dataset_names=[finalized_dataset],
+        model_names=[finalized_model],
+        offset=1,
+        limit=1,
+    )
+
+    assert len(evaluations) == 1
```
czaloom, line 450: Can you add a value check to show that the output is deterministic?

api/tests/functional-tests/backend/core/test_evaluation.py
```diff
@@ -421,13 +421,47 @@ def test_get_evaluations(
...
+    # check that having too high of a limit param doesn't throw an error
+    evaluations, headers = core.get_evaluations(
+        db=db,
+        dataset_names=[finalized_dataset],
+        model_names=[finalized_model],
+        offset=0,
+        limit=6,
+    )
+
+    assert len(evaluations) == 2
```
czaloom, line 462: Can you add a value check to show that the output is deterministic?

api/tests/functional-tests/backend/core/test_label.py
```diff
@@ -281,6 +282,34 @@ def test_get_labels(
...
+        schemas.Label(key="k1", value="v2"),
+        schemas.Label(key="k1", value="v3"),
+        schemas.Label(key="k3", value="v3"),
+    }
+    assert headers == {"content-range": "items 0-4/5"}
+
+    pred_labels, headers = get_paginated_labels(
+        db, ignore_groundtruths=True, offset=1, limit=100
+    )
+    assert len(pred_labels) == 2
```
czaloom, line 304: Can you add a value check to show that the output is deterministic?

api/valor_api/backend/core/label.py
czaloom: I would prefer that API endpoint related code (like pagination) be implemented in CRUD rather than the backend.

This would work by augmenting the getters in `_read.py` which are meant to be the interface between the API and the backend.

api/valor_api/backend/core/model.py
czaloom, line 1: See above. I feel that pagination makes more sense as a CRUD operation.

czaloom (changes requested):

api/valor_api/backend/core/model.py
czaloom: I think this applies to all the table getters.

api/tests/functional-tests/backend/core/test_evaluation.py
ntlind: none of the tests here check values because there isn't really anything deterministic to check. here's an example output from this breakpoint:

```python
# print(evaluations)

[EvaluationResponse(id=188, model_name='test_model', datum_filter=Filter(dataset_names=['test_dataset'], dataset_metadata=None, model_names=None, model_metadata=None, datum_uids=None, datum_metadata=None, task_types=None, annotation_metadata=None, require_bounding_box=None, bounding_box_area=None, require_polygon=None, polygon_area=None, require_raster=None, raster_area=None, labels=None, label_ids=None, label_keys=None, label_scores=None), parameters=EvaluationParameters(task_type=<TaskType.CLASSIFICATION: 'classification'>, convert_annotations_to_type=None, iou_thresholds_to_compute=None, iou_thresholds_to_return=None, label_map=None, recall_score_threshold=0.0, compute_pr_curves=None, pr_curve_iou_threshold=0.5), status=<EvaluationStatus.PENDING: 'pending'>, created_at=datetime.datetime(2024, 4, 12, 15, 30, 18, 963293, tzinfo=datetime.timezone.utc), metrics=[], confusion_matrices=[], ignored_pred_labels=None, missing_pred_labels=None)]
```

the id changes each run. let me know what you'd want to assert from the above

api/tests/functional-tests/backend/core/test_evaluation.py
ntlind: see above; not sure what value you'd want to check

api/valor_api/backend/core/label.py
ntlind: responded via slack. we want to use offset and limit in psql for performance reasons, and don't want to make db calls in crud functions

api/tests/functional-tests/backend/core/test_label.py
ntlind: added!

ntlind: this test isn't deterministic, unfortunately, because the `created_at` times vary across runs, and are often the same for multiple items. this means the sort order will vary across runs, and hence the output here will vary across runs. here's an example of `label` from one run:

image

ntlind: we could sort by key / value in addition to created_at if we wanted to, but doing so would introduce a small performance penalty for little added value IMO

czaloom: The value itself isn't what needs to be tested, I guess what im trying to say is the following:

```
labels_set == 
get_paginated_labels(
        db, ignore_groundtruths=True, offset=1, limit=100
    ) + get_paginated_labels(
        db, ignore_groundtruths=True, offset=0, limit=1
    )
```

to show that the queries are consistent.

czaloom: This type of test should exist for each implementation of pagination so that you can be certain that arbitrary offsets and limits will recombine to the same set.

api/tests/functional-tests/backend/core/test_evaluation.py
czaloom: See my comment below. I feel there needs to be a check that you actually get back all the same data if you combine multiple pages.

api/tests/functional-tests/backend/core/test_label.py
ntlind: good idea!

czaloom (approved):

api/valor_api/backend/core/label.py
```diff
@@ -320,6 +320,80 @@ def get_labels(
...
+    }
+
+    if labels:
+        end_index = (
+            offset + len(labels) - 1
+        )  # subtract one to make it zero-indexed
+
+        range_indicator = f"{offset}-{end_index}"
+    else:
+        range_indicator = "*"
```
rsbowman-striveworks, line 390: This `range_indicator` setting code appears in several places, might make sense to try to factor out.

ntlind: good idea! refactored this part into api_utils

rsbowman-striveworks (approved):
Nice, thanks!

PR #536, opened 2024-04-12
Rsb migrations fix

jyono:


jyono (commented):
this lgtm but do we still need it? I think no

rsbowman-striveworks:
Done in https://github.com/Striveworks/valor/pull/535 and https://github.com/Striveworks/valor/pull/538

PR #540, opened 2024-04-16
Add accuracy and true negative count to PR curves

ntlind:
## Improvements

Adds the count of true negatives and the accuracy score to each PR curve output for classification tasks. These two values will always be `None` for object detection tasks.


api/tests/functional-tests/backend/metrics/test_classification.py
```diff
@@ -919,35 +920,58 @@ def test__compute_curves(
...
         # cat
         ("cat", 0.05, "tp"): 1,
+        ("cat", 0.05, "tn"): 0,
         ("cat", 0.05, "fp"): 5,
         ("cat", 0.05, "fn"): 0,
         ("cat", 0.95, "tp"): 1,
         ("cat", 0.95, "fp"): 0,
+        ("cat", 0.95, "tn"): 5,
         ("cat", 0.95, "fn"): 0,
     }
```
czaloom, line 976: I may be missing something but for a specific label shouldnt there be a consistent total across score thresholds?

e.g. (TP + TN + FP + FN) @ 0.05 should equal (TP + TN + FP + FN) @ 0.95

For the label `bird` the total is equivalent for thresholds 0.05 and 0.65 but different for 0.3.

ntlind: what you wrote is correct; (TP + TN + FP + FN) should always add up to 6.

that's still the case for ('bird', 0.3), I'm just choosing not to show all four of the counts above (there's no real point, so they'll always add up to the same number). `len(curves['bird', 0.3, tn])` is 3, so the sum is still 6.

ntlind: update: added more checks so they all add up to 6

czaloom (approved):

PR #543, opened 2024-04-17
moved label key validation outside of computation

czaloom:
### Problem Description

Validation for classification tasks is silently failing when there is a label mismatch.

### Feature Description

Move validation outside of the computation.

### DIFF
--- a/api/valor_api/backend/core/evaluation.py
+++ b/api/valor_api/backend/core/evaluation.py
@@ -11,6 +11,35 @@ from valor_api.backend import core, models
 from valor_api.backend.query import Query
 
 
+def _validate_classification_task(
+    db: Session,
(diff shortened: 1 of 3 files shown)


api/valor_api/backend/core/evaluation.py
```diff
@@ -448,6 +477,14 @@ def create_or_get_evaluations(
                 parameters=subrequest.parameters.model_dump(),
                 status=enums.EvaluationStatus.PENDING,
             )
+
+            # validation
```
ntlind, line 481: nit: this comment doesn't add useful context

ntlind (approved):

rsbowman-striveworks (approved):

PR #546, opened 2024-04-18
Add basic analytics to the evaluation table

ntlind:
## Improvements

Stores the number of processed labels, datums, and annotations in a new `meta` field of the `evaluation` table. Also calculates and stores the duration (in seconds) that the evaluation took to run. Example output:

```python
...
eval_job = model.evaluate_classification(
    dataset, label_map=label_mapping, compute_pr_curves=True
)

print(eval_job.meta)

# {
#     "labels": 4, # the number of labels that were processed using the prescribed ground truth and prediction filters
#     "annotations": 6,
#     "datums": 2,
#     "duration": .34 # in seconds
# }

```

Right now, I'm assuming that this `meta` field will only store evaluation analytics. In the future, we might consider allowing users to attach their own metadata to an evaluation.


api/valor_api/backend/metrics/metric_utils.py
```diff
@@ -253,6 +256,109 @@ def create_metric_mappings(
...
+            gt_subquery.c.annotation_id,
+            gt_subquery.c.label_id,
+        ).select_from(gt_subquery)
+    ).all()
+
+    # handle edge case where no gts come back
+    if not gts:
+        gt_annotation_id, gt_label_id = set(), set()
+    else:
+        gt_annotation_id, gt_label_id = map(set, zip(*gts))
```
ntlind, line 300: Charles: does this process make sense to you? I was trying to minimize the number of psql calls (relative to querying the number of datums separate from the number of labels, etc.)

api/valor_api/backend/metrics/classification.py
```diff
@@ -930,4 +931,11 @@ def compute_clf_metrics(
             columns_to_ignore=["value"],
         )
 
+    log_evaluation_analytics(
```
czaloom, line 934: I think would prefer the call to this log creation at job creation not computation time. Open to your thoughts on this!

api/valor_api/backend/metrics/metric_utils.py
czaloom: Im not sure if this is possible, but the following would reduce the queries to 2.

```
db.query(
    Query(
        func.count(distinct(models.Datum.id)),
        func.count(distinct(models.Annotation.id)),
        func.count(distinct(models.Label.id)),
    )
    .filter(groundtruth_filter)
    .groundtruth()
).all()
```

api/valor_api/backend/metrics/segmentation.py
```diff
@@ -265,4 +266,11 @@ def compute_semantic_segmentation_metrics(
             columns_to_ignore=["value"],
         )
 
+    log_evaluation_analytics(
```
czaloom, line 269: see above

api/valor_api/backend/metrics/metric_utils.py
```diff
@@ -253,6 +256,109 @@ def create_metric_mappings(
...
+        Query(models.Dataset.name, models.Datum.id)  # type: ignore - sqlalchemy issues
+        .filter(groundtruth_filter)
+        .groundtruths()
+    ).all()
+    unique_datums = set(pd_datums + gt_datums)
+    datum_cnt = len(unique_datums)
+
+    evaluation = core.fetch_evaluation_from_id(db, evaluation_id)
+
+    server_time = db.execute(func.now()).scalar().replace(tzinfo=None)  # type: ignore - guaranteed to return server time if psql is running
```
czaloom, line 342: Why do we need this if an evaluation already has a `created_at` timestamp?

api/valor_api/backend/metrics/classification.py
czaloom: This could follow the pattern that were starting of moving eval validators to `backend/core/evaluation.py`

api/valor_api/backend/metrics/metric_utils.py
ntlind: to calculate duration, we need to compare the server time at completion to the `created_at` time

api/valor_api/backend/metrics/classification.py
ntlind: discussed why this isn't preferable via slack

api/valor_api/backend/metrics/metric_utils.py
```diff
@@ -253,6 +256,126 @@ def create_metric_mappings(
...
+        .alias()
+    )
+
+    gts = db.execute(
+        select(
+            gt_subquery.c.datum_id,
+            gt_subquery.c.annotation_id,
+            gt_subquery.c.label_id,
+        ).select_from(gt_subquery)
+    ).all()
```
czaloom, line 328: I thought i had left a comment about this but why cant we do something like this 

```
db.query(
    Query(
        func.count(distinct(models.Datum.id)),
        func.count(distinct(models.Annotation.id)),
        func.count(distinct(models.Label.id)),
    )
    .filter(groundtruth_filter)
    .groundtruths()
).all()
```

This would be more performant than counting in python

czaloom (approved):

PR #565, opened 2024-04-24
add check migrations workflow

ekorman:
This adds a GitHub actions workflow that checks if any migration files were altered and, if so, fails. 

Test cases: 

1. workflow on last commit, which has no changes to migrations folder: https://github.com/Striveworks/valor/actions/runs/8822524129/job/24220920820?pr=565
2. failure on a commit that altered an existing migrations file: https://github.com/Striveworks/valor/actions/runs/8822499639
3. passes on a commit that added a new migrations file: https://github.com/Striveworks/valor/actions/runs/8822517550


czaloom (dismissed):

.github/workflows/check-migrations.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - run: git fetch origin main
+      - name: Check for changed SQL migration files
+        run: |
+          # Compare current branch against main branch for modified files
+          CHANGED_FILES=$(git diff --name-status origin/main...HEAD migrations/sql | grep '^M' | cut -f2)
```
czaloom, line 19: This won't catch the condition where a developer deletes a migration file.

The migration tool also doesn't care about the name of the file only the prepended number. A user could delete `0...01_nameA.up.sql` and replace it with `0...01_nameB.up.sql` and this will pass.

ekorman: good point on deltion. do you want to allow renamed ones to pass or not?

ekorman: good point, just made the change to fail on deletes and renames

czaloom: This might work?

`git diff --name-status origin/main...HEAD api | grep -E "^M|^D" | cut -f2`

czaloom (approved):

ntlind (approved):

jyono (approved):
lgtm

Issue #482, opened 2024-03-13
ENH: finegrained exceptions in the client


ekorman:
### Feature Type

- [X] Adding new functionality to valor

- [ ] Changing existing functionality in valor

- [ ] Removing existing functionality in valor

### Problem Description

Right now we basically have a catch-all exception type `ClientException`. however, to support easier exception handling we should have client exceptions in one-to-one correspondence to the exceptions in `valor_api/exceptions.py` (except for `ServiceUnavailable`). While these can currently be checked by e.g. 

```python
try:
    ...
except ClientException as e:
    if e.detail["name"] == "DatumDoesNotExistError":
        # handle this exception
    else:
         raise e
```
this is not a good pattern

### Feature Description

Add exceptions in the client for every exception type in `valor_api.exceptions` (and these should all subclass `ClientException`). e.g. the above example would turn into

```python
try:
    ...
except DatumDoesNotExistError as e:
    # handle this exception
```

PR #567, opened 2024-04-26
add finegrained exceptions to client

ekorman:
Beyond adding finegrained exceptions to the client as outlined in #482 , this PR

- [ ] Removes some backend exceptions that were never used
- [ ] Changes behavior so that when calls like `get_dataset` or `get_prediction` are requested on things that don't exist, instead of returning `None` we raise an exception

### DIFF
--- a/client/valor/exceptions.py
+++ b/client/valor/exceptions.py
@@ -26,58 +26,107 @@ class ClientConnectionFailed(Exception):
 
 
 class ServiceUnavailable(ClientException):
+    """
+    Raises an exception if the Valor service is unavailble.
+    """
+
     pass
 
 
 class DatasetAlreadyExistsError(ClientException):
+    """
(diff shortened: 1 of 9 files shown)


client/valor/exceptions.py
```diff
@@ -18,3 +23,77 @@ def __init__(self):
 class ClientConnectionFailed(Exception):
     def __init__(self, msg: str):
         super().__init__(msg)
+
+
+class ServiceUnavailable(ClientException):
```
ntlind, line 28: would it make sense to copy the docstrings from the `/api` exception classes to these classes for clarity? The docstring for this one would be "Raises an exception if the Valor service is unavailable."

ntlind (approved):

client/valor/exceptions.py
ekorman: ya, will add

PR #569, opened 2024-04-29
Add text generation task type

bnativi:
## Improvements
- Added a new `evaluate_text_generation` task type that calculates four new metrics. These include text comparison metrics, which compare a prediction string to a groundtruth string. These also include llm-guided metrics which only sometimes require a groundtruth. The metrics are:
    - `AnswerRelevance` (Q&A, llm-guided)
    - `Bias` (general text generation, llm-guided)
    - `BLEU` (text comparison)
    - `Coherence` (general text generation, llm-guided)
    - `ContextRelevance` (RAG, llm-guided)
    - `Faithfulness` (RAG, llm-guided)
    - `Hallucination` (RAG, llm-guided)
    - `ROUGE` (text comparison)
    - `Toxicity` (general text generation, llm-guided)
-  Added text generation notebook with three example use cases (RAG, summarization and content generation).
- Added WrappedOpenAIClient and WrappedMistralAIClient to handle llm calls and llm-guided metric computations.
- Changed from alpine to slim.

## Testing
- Added API functional and unit tests for the text generation metrics and llm clients.
- Added client side integration tests for text generation metrics.
- Added external integration tests to test the llm-guided metrics with OpenAI's API and Mistral's API. Because those API's do not give us fully deterministic control, the integration tests only check that valid metrics are returned and do not check the exact metric values. 
    - These should only run on merge to main, and not on pushes to other branches.
 tests don't run when not merging to main
    - They pass when run.
pass on github when OPENAI_API_KEY is set
- If I purposely try to make them fail, say by setting the OPENAI_API_KEY to "", then they fail as expected. 
fail on github when OPENAI_API_KEY is not set
    - The secret API keys should only be available to the external API integration tests and not to the rest of the integration tests. When I added a test to integration_tests/client/ that tries to make non-mocked OPENAI API calls, the test fails because the api key is not available for this test. 
regular integration tests fail when they try to make non-mocked OpenAI API calls


api/valor_api/backend/metrics/llm_call.py
```diff
@@ -0,0 +1,277 @@
+from typing import Any
+
+import openai
+from mistralai.client import MistralAPIException, MistralClient
+from mistralai.models.chat_completion import ChatMessage
+
+COHERENCE_INSTRUCTION = """You are a helpful assistant. You will grade the user's text. Your task is to rate the text based on its coherence. Please make sure you read and understand these instructions carefully. Please keep this document open while reviewing, and refer to it as needed.
```
bnativi, line 7: I should cite deepeval for this instruction.

bnativi: This is the example from Deepeval's paper

You will be given one summary written for a news article.
Your task is to rate the summary on one metric.
Please make sure you read and understand these instructions carefully. Please keep this
document open while reviewing, and refer to it as needed.
Evaluation Criteria:
Coherence (1-5) - the collective quality of all sentences. We align this dimension with
the DUC quality question of structure and coherence whereby ”the summary should be
well-structured and well-organized. The summary should not just be a heap of related information, but should build from sentence to sentence to a coherent body of information about a topic.”
Evaluation Steps:
1. Read the news article carefully and identify the main topic and key points.
2. Read the summary and compare it to the news article. Check if the summary covers the main
topic and key points of the news article, and if it presents them in a clear and logical order.
3. Assign a score for coherence on a scale of 1 to 5, where 1 is the lowest and 5 is the highest
based on the Evaluation Criteria.
Example:
Source Text:
{{Document}}
Summary:
{{Summary}}
Evaluation Form (scores ONLY):
- Coherence:

api/valor_api/backend/metrics/llm_call.py
```diff
@@ -0,0 +1,277 @@
...
+Evaluation Criteria:
+Coherence (1-5) - the collective quality of all sentences. We align this dimension with the DUC quality question of structure and coherence whereby ”the summary should be well-structured and well-organized. The summary should not just be a heap of related information, but should build from sentence to sentence to a coherent body of information about a topic.”
+Evaluation Steps:
+1. Read the text carefully and identify the main topic and key points.
+2. Assign a score for coherence on a scale of 1 to 5, where 1 is the lowest and 5 is the highest based on the Evaluation Criteria. Respond with just the number 1 to 5."""
```
bnativi, line 12: I should cite Deepeval for this instruction. See appendix A of https://arxiv.org/pdf/2303.16634 for where I got this instruction from. 

However I also significantly modified their instruction. Deepeval mentions in their paper that the coherence metric should measure how coherently the output (Prediction.Annotation.text) follows from the input (Datum.text). However the definition of coherence you see here (which is exactly copied from an example in Deepeval's paper) sounds pretty different. It makes no mention of measuring the coherence of just the output, and instead defines coherence as "the collective quality of all sentences". 

The Deepeval example instruction instead uses the evaluation steps to indicate to the LLM that it should be judging the output text. However the evaluation steps in the example are overly specific (specifically designed for news article summarization). Deepeval deals with this by prompting an LLM to come up with evaluation steps for the given task (an approach they call G-Eval). We could take the same approach, however I wanted to do something simpler and easier for the user, at least for now, so I decided to simplify the evaluation steps and not include the input (Datum.text) in the LLM prompt. 

I think my approach seems better to Valor, as G-Eval seems overly involved compared to other metrics in Valor, however let me know if you think we should stick closer to Deepeval's approach.

bnativi: This is the example from Deepeval's paper

You will be given one summary written for a news article.
Your task is to rate the summary on one metric.
Please make sure you read and understand these instructions carefully. Please keep this
document open while reviewing, and refer to it as needed.
Evaluation Criteria:
Coherence (1-5) - the collective quality of all sentences. We align this dimension with
the DUC quality question of structure and coherence whereby ”the summary should be
well-structured and well-organized. The summary should not just be a heap of related information, but should build from sentence to sentence to a coherent body of information about a topic.”
Evaluation Steps:

Read the news article carefully and identify the main topic and key points.
Read the summary and compare it to the news article. Check if the summary covers the main
topic and key points of the news article, and if it presents them in a clear and logical order.
Assign a score for coherence on a scale of 1 to 5, where 1 is the lowest and 5 is the highest
based on the Evaluation Criteria.
Example:
Source Text:
{{Document}}
Summary:
{{Summary}}
Evaluation Form (scores ONLY):
Coherence:

api/valor_api/backend/metrics/llm_call.py
```diff
@@ -0,0 +1,277 @@
...
+    api_key: str | None = None
+    model_name: str
+
+    def __init__(
+        self,
+        api_key: str | None = None,
+        model_name: str | None = None,
+    ):
+        """
+        TODO Should we use an __attrs_post_init__ instead?
```
bnativi, line 36: I am not sure in what situations one would prefer __attrs_post_init__ over __init__

examples/text-generation/text_generation.ipynb
```diff
@@ -0,0 +1,1263 @@
+{
```
bnativi, line 1: Should I restrict the scope of this example notebook to just the metrics included in this PR for now?

ntlind: I would, but I think it depends how soon after this PR you plan to implement the rest of the metrics. if it's your next priority (ETA 1-2 weeks after you merge this PR), then it's probably fine to leave them

examples/text-generation/text_generation.ipynb
```diff
@@ -0,0 +1,1263 @@
...
+    "\n",
+    "The first example is RAG for Q&A. We get a RAG dataset from HuggingFace, use Llama-Index and GPT3.5-turbo to generate answers, and evaluate those answers with text comparison metrics, RAG metrics and general text generation metrics.\n",
+    "\n",
+    "The second example is Summarization. We download a CNN news dataset from HuggingFace which includes groundtruth summaries. We ask GPT3.5-turbo to summarize those articles. Then we evaluate those summaries with text comparison metrics, summarization metrics and general text generation metrics.\n",
+    "\n",
+    "The third example is content generation. We manually create a few queries, each of a different query type (creative, education, professional). Then we evaluate the generated content with general text generation metrics. \n",
+    "\n",
+    "TODO potentially change this text depending on our implementations of Naturalness, Coherence, and other metrics. "
```
ntlind, line 21: fix TODOs

integration_tests/client/datatype/test_data_generation.py
```diff
@@ -404,6 +404,7 @@ def test_generate_prediction_data(client: Client):
         },
         "parameters": {
             "task_type": TaskType.OBJECT_DETECTION.value,
+            "metric_params": None,
```
ntlind, line 407: note: I like this `metric_params` concept. if Charles/Eric like this idea, we could move a bunch of our other parameters under this umbrella.

integration_tests/client/metrics/test_text_generation.py
```diff
@@ -0,0 +1,405 @@
...
+    Client,
+    Dataset,
+    Datum,
+    GroundTruth,
+    Model,
+    Prediction,
+)
+from valor.enums import EvaluationStatus, MetricType
+
+PREDICTIONS = [
```
ntlind, line 21: I wish there was a bit more test variety. as it stands now, both functional tests and these integration tests all use the same examples

integration_tests/client/metrics/test_text_generation.py
```diff
@@ -0,0 +1,405 @@
...
+        # MetricType.Summarization,
+        # MetricType.Toxicity,
+    ]
+
+    eval_job = model.evaluate_text_generation(
+        datasets=dataset,
+        metrics_to_return=metrics_to_return,
+        llm_api_params={
+            "client": "openai",
+            # "api_url":"https://api.openai.com/v1/chat/completions",
```
ntlind, line 233: remove commented code?

integration_tests/client/metrics/test_text_generation.py
```diff
@@ -0,0 +1,405 @@
...
+                    ],
+                )
+            ],
+        ),
```
ntlind, line 404: add one test which uses a real API call (after getting the right tokens into GitHub)?

api/tests/functional-tests/backend/metrics/test_text_generation.py
```diff
@@ -0,0 +1,395 @@
...
+    # default request
+    job_request = schemas.EvaluationRequest(
+        dataset_names=[dataset_name],
+        model_names=[model_name],
+        parameters=schemas.EvaluationParameters(
+            task_type=enums.TaskType.TEXT_GENERATION,
+            metrics_to_return=metrics_to_return,
+            llm_api_params={
+                "client": "openai",
+                # "api_url": "https://api.openai.com/v1/chat/completions",
```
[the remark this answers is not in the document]
bnativi: Currently, the logic only works with OPENAI_API_KEY or MISTRAL_API_KEY. I added logic to allow the api_key to also be passed in llm_api_params, but don't have a test for that logic

api/valor_api/backend/metrics/llm_call.py
```diff
@@ -0,0 +1,266 @@
...
+        self,
+    ):
+        """
+        Setup the connection to the API.
+        """
+        raise NotImplementedError
+
+    def process_messages(
+        self,
+        messages: list[dict[str, str]],
```
bnativi, line 52: Should we use stricter validation here? messages should be in the OpenAI format, which means that the keys for the dictionaries should be "role" and "content". What would be the best way to validate this?

api/valor_api/backend/metrics/llm_call.py
bnativi: I modified this to a function. 

I am concerned though that I have strayed too far from the conventional use of coherence.

api/valor_api/backend/metrics/llm_call.py
```diff
@@ -0,0 +1,266 @@
...
+        messages = [
+            {"role": "system", "content": COHERENCE_INSTRUCTION},
+            {"role": "user", "content": text},
+        ]
+
+        response = self(messages)
+
+        try:
+            # Valid responses: "5", "\n5", "5\n", "5.", " 5", "5 {explanation}", etc.
+            ret = int(response.strip()[0])
```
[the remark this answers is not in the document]
bnativi: I'm only taking the first character of the striped string, so it can't be a float.

But still to your point, if the model returns "3.8", then this code would just take the "3" and convert to the integer 3. That doesn't seem ideal. However I instructed the model to respond with an integer, so if it responds with a float, it's misbehaving anyways. Do we care if we are handling that misbehavior in a slightly non-ideal way?

api/valor_api/backend/metrics/llm_call.py
```diff
@@ -0,0 +1,266 @@
...
+    seed : int, optional
+        An optional seed can be provided to GPT to get deterministic results.
+    model_name : str
+        The model to use. Defaults to "gpt-3.5-turbo".
+    """
+
+    # url: str
+    api_key: str | None = None
+    seed: int | None = None
+    model_name: str = "gpt-3.5-turbo"  # gpt-3.5-turbo gpt-4-turbo gpt-4o
```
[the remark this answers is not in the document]
bnativi: The valid model names for each API changes as OpenAI or Mistral changes what models they offer. It seems like we can't do much validation of that here.

docs/metrics.md
```diff
@@ -282,4 +293,58 @@ print(detailed_evaluation)
...
\ No newline at end of file
+```
+
+## Text Generation Metrics
+
+## Q&A Metrics
+
+### Answer Relevance
+
+Ben TODO: I took this content from [this page](https://docs.ragas.io/en/latest/concepts/metrics/answer_relevance.html); feel free to edit.
```
ntlind, line 304: TODO review these new docs and add in missing content

api/tests/functional-tests/backend/core/test_llm_clients.py
```diff
@@ -0,0 +1,302 @@
...
+        "valor_api.backend.core.llm_clients.LLMClient.__call__",
+        _return_valid_answer_relevance_response,
+    )
+
+    assert 0.5 == client.answer_relevance("some query", "some answer")
+
+    # patch __call__ with an invalid response
+    monkeypatch.setattr(
+        "valor_api.backend.core.llm_clients.LLMClient.__call__",
+        _return_invalid_response,
```
bnativi, line 94: Add another test for if _generate_statements() works as desired but _generate_anwswer_relevance_verdicts() gets a bad llm call

api/tests/functional-tests/backend/core/test_llm_clients.py
```diff
@@ -0,0 +1,302 @@
...
+            ],
+            created=int(datetime.datetime.now().timestamp()),
+        )
+
+    client = WrappedOpenAIClient(api_key=None, model_name="model_name")
+
+    fake_message = [{"key": "value"}]
+
+    # with pytest.raises(openai.OpenAIError):
+    #     client.connect()
```
bnativi, line 185: Why should this client.connect() fail? Should it throw an error if OPENAI_API_KEY is set? It wouldn't fail for me locally.

api/tests/functional-tests/backend/core/test_llm_clients.py
```diff
@@ -0,0 +1,302 @@
...
+                prompt_tokens=2, total_tokens=4, completion_tokens=199
+            ),
+        )
+
+    client = WrappedMistralAIClient(api_key=None, model_name="model_name")
+
+    fake_message = [{"role": "role", "content": "content"}]
+
+    # with pytest.raises(MistralException):
+    #     client.connect()
```
bnativi, line 275: same as above, why should this fail?

api/tests/functional-tests/backend/core/test_llm_clients.py
ntlind: that's right. it fails for me (and on the GitHub runner) since we don't have the OPENAI_API_KEY set. you could remove this test if you'd like; agree that this isn't necessary

api/tests/functional-tests/backend/core/test_llm_clients.py
```diff
@@ -0,0 +1,302 @@
+import datetime
+from unittest.mock import MagicMock
+
+# import openai
```
ntlind, line 4: remove commented code?

api/valor_api/schemas/evaluation.py
```diff
@@ -116,6 +135,33 @@ def _validate_parameters(cls, values):
                             raise ValueError(
                                 "`iou_thresholds_to_return` must be a subset of `iou_thresholds_to_compute`"
                             )
+            case TaskType.TEXT_GENERATION:
+                text_comparison_metrics = set(["ROUGE", "BLEU"])
```
bnativi, line 139: metric validation needed for ROUGE and BLEU

api/tests/unit-tests/schemas/test_metrics.py
```diff
@@ -382,3 +382,127 @@ def test_mIOUMetric():
             for key in iou_metric.db_mapping(evaluation_id=1)
         ]
     )
+
+
```
bnativi, line 386: Need tests for AnswerRelevance and Coherence

api/valor_api/backend/core/annotation.py
```diff
@@ -77,6 +78,12 @@ def _create_annotation(
         raster = annotation.raster.to_psql()
     if annotation.embedding:
         embedding_id = _create_embedding(db=db, value=annotation.embedding)
+    if annotation.context:
```
bnativi, line 81: tests for str, list[str], invalid

api/valor_api/backend/core/llm_clients.py
```diff
@@ -0,0 +1,521 @@
...
+            assert type(response) == dict
+            assert "verdicts" in response
+            verdicts = response["verdicts"]
+            assert type(verdicts) == list
+            assert len(verdicts) == len(statements)
+            assert all(
+                verdict["verdict"] in ["yes", "no", "idk"]
+                for verdict in verdicts
+            )
+        except Exception:
```
bnativi, line 281: This looks messy and doesn't indicate to the user what the actual issue was with response. Also, the use of assert statements makes it unclear which fail conditions are actually being tested.

api/valor_api/backend/core/llm_clients.py
```diff
@@ -0,0 +1,521 @@
...
+            },
+        ]
+
+        response = self(messages)
+        try:
+            response = trim_and_load_json(response)
+            assert type(response) == dict
+            assert "statements" in response
+            statements = response["statements"]
+        except Exception:
```
bnativi, line 234: This looks messy and doesn't indicate to the user what the actual issue was with response. Also, the use of assert statements makes it unclear which fail conditions are actually being tested.

api/valor_api/backend/core/llm_clients.py
```diff
@@ -0,0 +1,521 @@
...
+            {"role": "user", "content": _get_coherence_instruction(text)},
+        ]
+
+        response = self(messages)
+
+        try:
+            # Valid responses: "5", "\n5", "5\n", "5.", " 5", "5 {explanation}", etc.
+            ret = int(response.strip()[0])
+            assert ret in {1, 2, 3, 4, 5}
+        except Exception:
```
bnativi, line 342: same comment as for answer relevance above. This might make it hard for the user to understand the issue with the llm's response

api/valor_api/backend/core/llm_clients.py
```diff
@@ -0,0 +1,521 @@
...
+
+    def connect(
+        self,
+    ):
+        """
+        Setup the connection to the API.
+        """
+        if self.api_key is None:
+            self.client = openai.OpenAI()
+        else:
```
bnativi, line 391: test that a manual key works

api/valor_api/schemas/evaluation.py
```diff
@@ -86,7 +92,20 @@ def _validate_parameters(cls, values):
                         MetricType.IOU,
                         MetricType.mIOU,
                     ]
+                case TaskType.TEXT_GENERATION:
```
bnativi, line 95: test for this error

client/valor/schemas/symbolic/types.py
```diff
@@ -2124,3 +2124,64 @@ def get_type_by_name(
         return List[type_]
     else:
         raise NotImplementedError(name)
+
+
+class Text(String):
```
bnativi, line 2129: Do we need tests for this and Context? I didn't see any tests for other classes in schemas/symbolic/types in the valor_client unit tests.

integration_tests/client/metrics/test_text_generation.py
bnativi: Discussion with Eric: we should discuss if we want to test actual calls to OpenAI and Mistral or not

integration_tests/client/metrics/test_text_generation.py
```diff
@@ -0,0 +1,478 @@
...
+
+@patch(
+    "valor.client.ClientConnection.evaluate",
+    mock_ClientConnection_evaluate,
+)
+@patch(
+    "valor.client.ClientConnection.get_evaluations",
+    mock_ClientConnection_get_evaluations,
+)
+def test_llm_evaluation(
```
bnativi, line 253: We should also test BLEU and ROUGE here.

api/valor_api/backend/metrics/text_generation.py
```diff
@@ -0,0 +1,526 @@
...
+                bleu_metrics = _calculate_sentence_bleu(
+                    predictions=predictions,
+                    references=references,
+                    weights=metric_params.get(
+                        "weights", [0.25, 0.25, 0.25, 0.25]
+                    ),
+                )
+
+                output += [
+                    schemas.BLEUMetric(
```
bnativi, line 352: @ntlind This metric does not make any mention of the groundtruth that it was compared to. Is this because it is compared to all groundtruths associated to the datum and the bleu score reported is the max?

api/Dockerfile
```diff
@@ -1,18 +1,17 @@
-FROM python:3.10-alpine
+FROM python:3.10-slim
 
-ARG VERSION="0.0.0-dev"
+ARG VALOR_VERSION="0.0.0-dev"
```
ekorman, line 3: what was the reason for this change? this also requires updating `.github/workflows/api-docker.yml` (line 68)

also since this doesn't seem related to this PR i prefer this be a separate one

api/valor_api/backend/metrics/text_generation.py
```diff
@@ -0,0 +1,526 @@
...
+                metrics[type_][i], output[prediction][type_]
+            )
+
+    return [
+        {"prediction": prediction, "value": dict(value)}
+        for prediction, value in output.items()
+    ]
+
+
+def _calculate_sentence_bleu(
```
bnativi, line 108: Add a smoothing function parameter

api/Dockerfile
ntlind: thanks for calling this out! this was leftover from when I was trying to debug the build errors using `python-alpine`; will revert this back

integration_tests/client/metrics/test_text_generation.py
```diff
@@ -0,0 +1,708 @@
...
+    """Yes, Lincoln won the election of 1860. He received the highest number of votes and a majority in the Electoral College, making him the 16th President of the United States. However, it's important to note that he won entirely due to his support in the North and West, as he did not receive any votes in 10 of the 15 Southern slave states.""",
+    """If a turtle egg was kept warm, it would likely hatch into a baby turtle. The sex of the baby turtle would be determined by the incubation temperature, assuming the species is one of those that determine sex thermally. This is because many turtle species have the ability to move around inside their eggs to select the best temperature for development, which can influence their sexual destiny.""",
+]
+
+
+def mock_ClientConnection_evaluate(
```
ekorman, line 28: why are we mocking this instead of using the real backend (which is the point of the integration tests)?

api/valor_api/backend/core/annotation.py
```diff
@@ -77,6 +78,12 @@ def _create_annotation(
...
+        if isinstance(annotation.context, str):
+            context = [annotation.context]
+        else:
+            assert isinstance(annotation.context, list)
```
czaloom, line 85: - `assert` should not be used in production code as it can be ignored depending on python interpreter flags. See [link](https://stackoverflow.com/questions/944592/best-practice-for-using-assert).
- maybe add `if not all([isinstance(item, str) for item in annotation.context])` to check that the contents of `annotaiton.context` is of type `str`

api/valor_api/backend/core/llm_clients.py
```diff
@@ -0,0 +1,641 @@
...
+            {"role": "system", "content": DEFAULT_SYSTEM_PROMPT},
+            {"role": "user", "content": _get_coherence_instruction(text)},
+        ]
+
+        response = self(messages)
+
+        try:
+            # Valid responses: "5", "\n5", "5\n", "5.", " 5", "5 {explanation}", etc.
+            ret = int(response.strip()[0])
+            assert ret in {1, 2, 3, 4, 5}
```
czaloom, line 366: `assert` should not be used in production code as it can be ignored depending on python interpreter flags. See [link](https://stackoverflow.com/questions/944592/best-practice-for-using-assert).

This should be changed into explicit exception, it may be good to create a custom exception in `valor_api.exceptions`. A possible exception could be `MalformedLLMResponseError` which you could pass the response into. This would inform the user what the model responded with.

api/valor_api/backend/metrics/text_generation.py
```diff
@@ -0,0 +1,529 @@
...
+
+    metrics = rouge.compute(
+        predictions=processed_prediction,
+        references=processed_references,
+        rouge_types=rouge_types,
+        use_stemmer=use_stemmer,
+        use_aggregator=False,  # aggregation gives us an average across all predicitons, which isn't what we want
+    )
+
+    assert metrics is not None  # handle type error
```
czaloom, line 93: `assert` should not be used in production code as it can be ignored depending on python interpreter flags. See [link](https://stackoverflow.com/questions/944592/best-practice-for-using-assert).

I would suggest changing this to `if metrics is None: raise TypeError(message)`

api/valor_api/backend/metrics/text_generation.py
czaloom: has this been completed?

api/valor_api/backend/metrics/text_generation.py
```diff
@@ -0,0 +1,529 @@
...
+                f"Client {llm_api_params['client']} is not supported."
+            )
+    else:
+        raise NotImplementedError(
+            "Support has not been implemented for api_url."
+        )
+
+    client_kwargs = {}
+    if "data" in llm_api_params:
+        assert isinstance(llm_api_params["data"], dict)
```
czaloom, line 221: `assert` should not be used in production code as it can be ignored depending on python interpreter flags. See [link](https://stackoverflow.com/questions/944592/best-practice-for-using-assert).

I would suggest changing this to `if not is_instance(..., dict): raise ...`

api/valor_api/backend/metrics/text_generation.py
```diff
@@ -0,0 +1,529 @@
...
+    parameters = schemas.EvaluationParameters(**evaluation.parameters)
+
+    log_evaluation_item_counts(
+        db=db,
+        evaluation=evaluation,
+        prediction_filter=prediction_filter,
+        groundtruth_filter=groundtruth_filter,
+    )
+
+    assert parameters.metrics_to_return
```
czaloom, line 492: `assert` should not be used in production code as it can be ignored depending on python interpreter flags. See [link](https://stackoverflow.com/questions/944592/best-practice-for-using-assert).

This should be changed to an explicit "if x then raise"

api/valor_api/schemas/evaluation.py
czaloom: has this been completed?

api/valor_api/schemas/evaluation.py
czaloom: has this been completed?

client/valor/schemas/symbolic/types.py
czaloom: For future reference, you can check this in test coverage in the github runner under  `combine-coverage-report`

client/valor/schemas/symbolic/collections.py
```diff
@@ -332,6 +344,8 @@ class Annotation(StaticCollection):
     implied_task_types: SymbolicList[String] = SymbolicList[String].symbolic(
         owner="annotation", name="implied_task_types"
     )
+    text: Text = Text.symbolic(owner="annotation", name="text")
+    context: Context = Context.symbolic(owner="annotation", name="context")
```
czaloom, line 348: Instead of defining Text and Context you could also have used 
- `String.nullable(owner="annotation", name="text")`
- `TypedList[String].nullable(owner="annotation", name="context")`

Im okay with leaving `Text` and `Context` if they provide other functionality.

client/valor/schemas/symbolic/types.py
czaloom: You may also be able to remove these objects by changing `Text` to `String.nullable` and `Context` to `TypedList[String].nullable` within the `Annotation` object.

integration_tests/client/metrics/test_text_generation.py
czaloom: has/will this be addressed in this pr?

integration_tests/client/metrics/test_text_generation.py
czaloom: has this been addressed?

api/valor_api/backend/core/annotation.py
czaloom: has this been addressed?

api/tests/unit-tests/schemas/test_metrics.py
czaloom: has this been addressed?

czaloom (changes requested):

integration_tests/client/metrics/test_text_generation.py
bnativi: This should be addressed in future PRs as we need different data for summarization metrics.

api/tests/functional-tests/backend/core/test_llm_clients.py
```diff
@@ -0,0 +1,488 @@
...
+                    ),
+                )
+            ],
+            created=int(datetime.datetime.now().timestamp()),
+            usage=UsageInfo(
+                prompt_tokens=2, total_tokens=4, completion_tokens=199
+            ),
+        )
+
+    def create_mock_chat_completion(model, messages) -> ChatCompletionResponse:
```
ntlind, line 370: nit: all of these functions could probably start with an underscore to indicate that they are meant to be private

api/tests/functional-tests/backend/core/test_llm_clients.py
```diff
@@ -0,0 +1,488 @@
...
+
+    # Needs to have 'verdicts' key.
+    monkeypatch.setattr(
+        "valor_api.backend.core.llm_clients.LLMClient.__call__",
+        _return_invalid3_answer_relevance_response,
+    )
+    with pytest.raises(InvalidLLMResponseError):
+        client.answer_relevance("some query", "some text")
+
+    # Invalid verdict.
```
ntlind, line 202: super nitty suggestions; feel free to not make these changes: 
- some comments are proper sentences with periods, while others aren't (e.g., "# patch __call__ with a valid response"). it'd be ideal if they were all consistent
- overall there's a lot of comments here, some of which are super helpful for understanding your intent (e.g., "Coherence score was 0, which is not in {1,2,3,4,5}") while others probably aren't necessary since the code itself makes it obvious what you're trying to do (e.g., "# patch __call__ with a valid response")

api/tests/functional-tests/backend/metrics/test_text_generation.py
```diff
@@ -0,0 +1,957 @@
...
+    """If kept warm, it would hatch a coyote.""",  # very different than prediction
+]
+
+
+def mocked_connection(self):
+    pass
+
+
+def mocked_answer_relevance(self, query: str, text: str):
+    if text in [PREDICTIONS[0]]:
```
ntlind, line 35: nit: using a dictionary here probably makes sense vs. if blocks

api/tests/functional-tests/backend/metrics/test_text_generation.py
```diff
@@ -0,0 +1,957 @@
...
+from valor_api.backend.core import create_or_get_evaluations
+from valor_api.backend.metrics.text_generation import (
+    _calculate_rouge_scores,
+    _calculate_sentence_bleu,
+    _compute_text_generation_metrics,
+    compute_text_generation_metrics,
+)
+
+# Moved PREDICTIONS here so that they are accessible by mock functions.
+PREDICTIONS = [
```
ntlind, line 17: nit: should PREDICTIONS and REFERENCES be fixtures (similar to text_generation_test_data)? they'd still be accessible by mock functions

api/tests/unit-tests/schemas/test_evaluation.py
```diff
@@ -4,6 +4,15 @@
 from pydantic import ValidationError
 
 from valor_api import enums, schemas
+from valor_api.enums import MetricType
+
+LLM_API_PARAMS = {
```
ntlind, line 9: nit: could this be a fixture instead of a module-level variable?

api/valor_api/backend/core/llm_clients.py
```diff
@@ -0,0 +1,659 @@
...
+JSON:
+"""
+
+
+def _get_coherence_instruction(text: str) -> str:
+    """
+    This instruction was adapted from appendix A of DeepEval's paper G-EVAL: NLG Evaluation using GPT-4 with Better Human Alignment (https://arxiv.org/pdf/2303.16634).
+    The main adaptation is a generalization of the metric to more task types. The example prompt in DeepEval was specific to summarization, but the below prompt could apply to any text generation task.
+    Crucially, unlike DeepEval, no context is used. Instead, the coherence of the text is evaluated entirely based on the text. This generalizes the prompt and also prevents the evaluation from being influenced by the quality of sentences in the context.
+    Parameters
```
ntlind, line 124: nit: add a newline here between the text and "Parameters"? can't tell if one exists or not

api/valor_api/backend/core/llm_clients.py
```diff
@@ -0,0 +1,659 @@
...
+    ):
+        """
+        Setup the connection to the API. Not implemented for parent class.
+        """
+        raise NotImplementedError
+
+    def process_messages(
+        self,
+        messages: list[dict[str, str]],
+    ) -> Any:
```
ntlind, line 189: should this really be Any?

api/valor_api/backend/core/prediction.py
```diff
@@ -127,12 +127,21 @@ def create_predictions(
         predictions, annotation_ids
     ):
         for i, annotation in enumerate(prediction.annotations):
-            for label in annotation.labels:
+            if annotation.labels:
```
ntlind, line 130: is this tested anywhere?

api/valor_api/backend/metrics/text_generation.py
```diff
@@ -0,0 +1,555 @@
...
+    metrics = rouge.compute(
+        predictions=processed_prediction,
+        references=processed_references,
+        rouge_types=rouge_types,
+        use_stemmer=use_stemmer,
+        use_aggregator=False,  # aggregation gives us an average across all predicitons, which isn't what we want
+    )
+
+    if not metrics:
+        raise ValueError("No metrics were returned.")
```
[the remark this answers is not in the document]
ntlind: I'm pretty sure I added this to handle a type error where `rouge.compute` can return None. I don't have an issue with you deleting it if you don't think it's necessary

api/valor_api/backend/metrics/text_generation.py
ntlind: @bnativi did you add tickets for enhancements like this? they'll be useful after this PR when trying to prioritize what comes next

api/valor_api/schemas/types.py
```diff
@@ -85,20 +90,34 @@ def _match_annotation_to_implied_task_type(
...
         and annotation.polygon is None
         and annotation.raster is None
+        and annotation.text is None
+        and annotation.context is None
     ):
         implied_type = ["empty"]
     else:
         raise ValueError(
-            "Input didn't match any known patterns. Classification tasks should only contain labels. Object detection tasks should contain labels and polygons, bounding boxes, or rasters with is_instance == True. Segmentation tasks should contain labels and rasters with is_instance != True."
+            "Input didn't match any known patterns. Classification tasks should only contain labels. Object detection tasks should contain labels and polygons, bounding boxes, or rasters with is_instance == True. Segmentation tasks should contain labels and rasters with is_instance != True. Text Generation tasks should only contain text and optionally context."
```
ntlind, line 120:
```suggestion
            "Input didn't match any known patterns. Classification tasks should only contain labels. Object detection tasks should contain labels and polygons, bounding boxes, or rasters with is_instance == True. Segmentation tasks should contain labels and rasters with is_instance != True. Text generation tasks should only contain text and, optionally, context."
```

api/valor_api/schemas/types.py
```diff
@@ -199,6 +218,9 @@ def _validate_prediction_annotations(annotations: list["Annotation"]) -> None:
                 labels.append(label)
                 indices[label] = index
 
+        elif "text-generation" in annotation.implied_task_types:
```
ntlind, line 221: is this line necessary?

api/valor_api/schemas/types.py
```diff
@@ -386,11 +416,14 @@ class Datum(BaseModel):
     ----------
     uid : str
         The UID of the datum.
+    text : str
```
ntlind, line 419: should this docstring indicate that this var is optional?

integration_tests/external/test_text_generation_no_mock.py
```diff
@@ -0,0 +1,266 @@
...
+
+    # Check that the returned metrics have the right format.
+    for m in metrics:
+        if m["type"] in ["AnswerRelevance", "BLEU"]:
+            assert 0 <= m["value"] <= 1
+        if m["type"] == "Coherence":
+            assert m["value"] in {1, 2, 3, 4, 5}
+        if m["type"] == "ROUGE":
+            assert isinstance(m["value"], dict)
+            assert all(0 <= v <= 1 for v in m["value"].values())
```
ntlind, line 88: can you check the values of what's returned? i.e., coherence value should be >= 4, for example

integration_tests/external/test_text_generation_no_mock.py
```diff
@@ -0,0 +1,266 @@
...
+                Annotation(text=REFERENCES[2]),
+                Annotation(text="some other text"),
+                Annotation(text="some final text"),
+            ],
+        ),
+    ]
+
+
+@pytest.fixture
+def pred_answers(
```
ntlind, line 218: move gt and pred fixtures to the top of the file for consistency?

api/valor_api/backend/core/llm_clients.py
czaloom: Second this, the mistral client below seems to return `list[ChatMessage]` while the OpenAI client uses this base function.

I think you can retain the `-> Any` if you make this base function `raise NotImplementedError` and move the contents to the OpenAI client.

If multiple clients use the default function and you want to preserve a single point of truth then maybe a middle class would be better. The inheritance would be `LLMClient --> OpenAIStyleClient  --> OpenAIClient`

client/valor/coretypes.py
```diff
@@ -39,9 +40,9 @@ class GroundTruth(StaticCollection):
 
     Attributes
     ----------
-    datum : Datum
```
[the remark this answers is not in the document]
czaloom: I would prefer to add back all these spaces per the[ numpy spec ](https://numpydoc.readthedocs.io/en/latest/format.html#parameters)

client/valor/schemas/evaluation.py
```diff
@@ -27,21 +29,31 @@ class EvaluationParameters:
...
         The confidence score threshold for use when determining whether to count a prediction as a true positive or not while calculating Average Recall.
     pr_curve_iou_threshold: float, optional
-            The IOU threshold to use when calculating precision-recall curves for object detection tasks. Defaults to 0.5.
+        The IOU threshold to use when calculating precision-recall curves for object detection tasks. Defaults to 0.5.
     pr_curve_max_examples: int
         The maximum number of datum examples to store when calculating PR curves.
+    bleu_weights: list[float], optional
+        The weights to use when calculating BLEU scores.
+    rouge_types: list[str]
+        A list of rouge types to calculate. Options are ['rouge1', 'rouge2', 'rougeL', 'rougeLsum'], where `rouge1` is unigram-based scoring, `rouge2` is bigram-based scoring, `rougeL` is scoring based on sentences (i.e., splitting on "." and ignoring "\n"), and `rougeLsum` is scoring based on splitting the text using "\n".
```
czaloom, line 38: Since these options are definite i would prefer this be a enum on both the client and the backend.

client/valor/schemas/evaluation.py
```diff
@@ -27,21 +29,31 @@ class EvaluationParameters:
...
+    llm_api_params: Optional[Dict[str, Union[str, dict]]] = None
 
     convert_annotations_to_type: Optional[AnnotationType] = None
     iou_thresholds_to_compute: Optional[List[float]] = None
     iou_thresholds_to_return: Optional[List[float]] = None
     recall_score_threshold: float = 0
     pr_curve_iou_threshold: float = 0.5
     pr_curve_max_examples: int = 1
+    bleu_weights: Optional[List[float]] = None
+    rouge_types: Optional[List[str]] = None
```
czaloom, line 55: Same as above comment, i would prefer `rouge_types` be a list of an enum

client/valor/coretypes.py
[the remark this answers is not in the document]
czaloom: Although it seems like weve been inconsistent with enforcing this up to this point

client/valor/coretypes.py
```diff
@@ -1032,13 +1033,120 @@ def evaluate_segmentation(
...
+                )
+                rouge_use_stemmer = metric_params[MetricType.ROUGE].get(
+                    "use_stemmer"
+                )
+
+        # format request
+        datasets = datasets if isinstance(datasets, list) else [datasets]
+        filters = filters if filters else Filter()
+        request = EvaluationRequest(
+            dataset_names=[dataset.name for dataset in datasets],  # type: ignore - issue #604
```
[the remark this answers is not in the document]
bnativi: I just left it that way because it is that way for the other 3 task types. Issue 604 isn't specific to text generation, so if this should be removed here, then it should be removed everywhere. 

Looks like #604 is @czaloom ?

api/valor_api/backend/core/prediction.py
bnativi: What do you want to test here? This code is all covered by tests. There isn’t a functional test or unit test for prediction.py, but we could add one.

client/valor/coretypes.py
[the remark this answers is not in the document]
bnativi: Is there a way to automatically standardize this? Otherwise, I can just change it back to the way it was.

api/tests/functional-tests/backend/metrics/test_text_generation.py
bnativi: When I change predictions to a fixture (as a function), it is accessible for the tests and other fixtures, but not for the mock functions. I'm adding predictions as a parameter of the mock functions. Is there another way to do this?

bnativi: I changed references to a fixture, but left PREDICTIONS as a module level variable, as it doesn't work as a fixture with mock functions.

client/valor/coretypes.py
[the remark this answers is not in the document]
bnativi: I reverted it in this file

integration_tests/external/test_text_generation_no_mock.py
bnativi: I broke this into two functions each, with data more curated to getting a good/bad metric score.

api/valor_api/backend/metrics/text_generation.py
```diff
@@ -0,0 +1,570 @@
...
+            .subquery()
+        )
+
+        joint_subquery = (
+            select(
+                groundtruth_subquery.c.datum_uid,
+                groundtruth_subquery.c.dataset_name,
+                functions.array_agg(
+                    prediction_subquery.c.prediction_text
+                ).label("predictions"),
```
bnativi, line 327: @ntlind Why do we array_agg here? If there are multiple predictions for one datum, then array_agg will combine them into a single list. That seems like that could break or at least degrade BLEU and ROUGE performance. It seems like we should be instead returning one BLEU score per prediction, not one BLEU score per datum.

bnativi: Oh I see, I didn't realize that the sentence bleu function below takes in the list of predictions but computes a separate score for each one

ntlind:
the code, tests, and notebook all look good to me. I'm ready to approve once we figure out why the benchmarks are now failing.

api/valor_api/backend/metrics/metric_utils.py
```diff
@@ -553,10 +569,46 @@ def prepare_filter_for_evaluation(
         )
     )
 
+    datum_filter = filters.model_copy()
+    datum_filter.groundtruths = None
+    datum_filter.predictions = None
```
czaloom, line 574: This datum filter is concerning to me as it could include datums outside what the user specifies.

To perform datum queries i would use the groundtruth_filter with `label_source=models.Annotation`

bnativi: @czaloom What if the datums don't have any labels?

czaloom: It wont remove them, it just sets the source if labels are specified.

api/valor_api/backend/models.py
```diff
@@ -73,7 +73,8 @@ class GroundTruth(Base):
         ForeignKey("annotation.id"), nullable=True
     )
     label_id: Mapped[int] = mapped_column(
-        ForeignKey("label.id"), nullable=False
+        ForeignKey("label.id"),
+        nullable=True,
```
czaloom, line 77: Please revert this changes as it does not affect your PR but may affect other evaluations. Since no labels are created, ground truths will also never be created.

api/valor_api/backend/models.py
```diff
@@ -99,7 +100,8 @@ class Prediction(Base):
...
```
czaloom, line 104: Please revert this changes as it does not affect your PR but may affect other evaluations. Since no labels are created, predictions will also never be created.

api/valor_api/backend/models.py
czaloom, line 77: This change confuses me as the GroundTruth table exists only to link labels. Since no other column was added, I think this should probably be reverted along with the migration change.

api/valor_api/backend/models.py
czaloom, line 104: This change also confuses me as the Prediction table exists primarily to link labels. Is text generation storing scores here?

api/valor_api/backend/models.py
bnativi: @czaloom There aren't any labels for text generation groundtruths and predictions, so there is no label id. You can see in api/valor_api/backend/core/prediction.py lines 161:175 for example that I had to save the Prediction with label_id = None because there aren't any labels.

api/valor_api/backend/models.py
bnativi: See my comment for groundtruths

api/valor_api/backend/models.py
czaloom: Are there scores involved? The ground truth and prediction tables in the backend are really only label linkers and should be created if there are no labels.

ntlind (approved):

Issue #571, opened 2024-04-30
BUG: Unexpected behavior filtering valor objects with list comprehensions 


jyono:
### valor version checks

- [X] I have confirmed this bug exists on the latest version of valor.

### Reproducible Example

```python
from valor import connect, Client, Dataset, Model, Datum, Annotation, GroundTruth, Prediction, Label
from valor.schemas import Box
from valor.enums import TaskType

connect("https://valor.striveworks.com/api/v1", username=..., password=...)
client = Client()

pscope_test_dataset = Dataset.get("PlanetScope 1.2 Test")

sample_gt = pscope_test_dataset.get_groundtruth('')
sample_annotations = sample_gt.annotations
sample_labels = sample_annotations[0].labels

print(f"sample_labels = {sample_labels}")
print(f"sample_labels[0].value == 'some' => {sample_labels[0].value == 'some'}")
print(f"sample_labels[0].value == 'thing' => {sample_labels[0].value == 'thing'}")
print(f"[label for label in sample_labels if label.value == 'some'] = {[label for label in sample_labels if label.value == 'some']}")
print(f"[label for label in sample_labels if str(label.value) == 'some'] = {[label for label in sample_labels if str(label.value) == 'some']}")
```

### Issue Description

I am getting unexpected behavior when filtering valor objects with list comprehensions, specifically that the filtering does not work unless I apply “str()” to the value to filter on. The following code snippet demonstrates the issue (“username” and “password” need to be provided to the connect() call):

```
from valor import connect, Client, Dataset, Model, Datum, Annotation, GroundTruth, Prediction, Label
from valor.schemas import Box
from valor.enums import TaskType

connect("https://valor.striveworks.com/api/v1", username=..., password=...)
client = Client()

pscope_test_dataset = Dataset.get("PlanetScope 1.2 Test")

sample_gt = pscope_test_dataset.get_groundtruth('')
sample_annotations = sample_gt.annotations
sample_labels = sample_annotations[0].labels

print(f"sample_labels = {sample_labels}")
print(f"sample_labels[0].value == 'some' => {sample_labels[0].value == 'some'}")
print(f"sample_labels[0].value == 'thing' => {sample_labels[0].value == 'thing'}")
print(f"[label for label in sample_labels if label.value == 'some'] = {[label for label in sample_labels if label.value == 'some']}")
print(f"[label for label in sample_labels if str(label.value) == 'some'] = {[label for label in sample_labels if str(label.value) == 'some']}")
```

output 
[...]

PR #572, opened 2024-04-30
expose values with getattr

ekorman:
this PR does the following to address https://github.com/Striveworks/valor/issues/571:

- for attributes of a `StaticCollection` that are simple types (e.g. `String`, `Integer`, etc), have `__getattribute__` return the value
- update `__repr__` to be explicit that these objects are not standard types. this is something that made me confused when i was using valor because it made me think that certain objects such as `String`, `Integer` were built-in types


ntlind:
#571

client/valor/coretypes.py
```diff
@@ -591,7 +588,7 @@ def get_datums(
             raise ValueError(
                 "Cannot filter by dataset_names when calling `Dataset.get_datums`."
             )
-        filter_["dataset_names"] = [self.get_name()]
+        filter_["dataset_names"] = [self.name]  # type: ignore
```
czaloom, line 591: Adding type ignores is why i had avoided this previously. This will affect IDE autocompletion for users that have pylance enabled.

client/valor/schemas/symbolic/collections.py
```diff
@@ -42,6 +43,9 @@ def __init__(self, **kwargs):
             raise ValueError(
                 f"Expected the following keyword arguments '{static_keys}'. Received '{kwarg_keys}'."
             )
+        # import pdb
+
+        # pdb.set_trace()
```
czaloom: remove tests

client/valor/schemas/symbolic/collections.py
```diff
@@ -135,8 +145,14 @@ def decode_value(cls, value: dict):
 
     def encode_value(self):
         """Encode object to JSON compatible dictionary."""
+        # TODO: make this less hacky
         return {
-            k: v.encode_value() for k, v in self._get_dynamic_values().items()
+            k: (
+                v.encode_value()
+                if v is not None and hasattr(v, "encode_value")
+                else v
```
czaloom, line 149: what condition causes this `v = None`? A `StaticCollection` should convert build-in types to symbolic types at initialization

czaloom (changes requested):

client/valor/schemas/symbolic/collections.py
czaloom: symbolic attributes can contain the value `None` but are not themselves nullable

client/valor/coretypes.py
ekorman: yea i think this is worth the tradeoff though

client/valor/schemas/symbolic/collections.py
ekorman: yea i think you're right, just dropped that check

client/valor/schemas/symbolic/collections.py
```diff
@@ -136,7 +143,8 @@ def decode_value(cls, value: dict):
     def encode_value(self):
         """Encode object to JSON compatible dictionary."""
         return {
-            k: v.encode_value() for k, v in self._get_dynamic_values().items()
+            k: (v.encode_value() if hasattr(v, "encode_value") else v)
+            for k, v in self._get_dynamic_values().items()
```
czaloom, line 147: I think this can be completely reverted to the original 

```
k: v.encode_value() for k, v in self._get_dynamic_values().items()
```

as all symbolic types have the encode_value member

czaloom (approved):

PR #574, opened 2024-05-04
optimize groundtruths and predictions creation

ekorman:
This optimizes the adding of groundtruths and predictions by

1. actually add them in the batches they were sent instead of one by one
2. use bulk adds with mappings instead of ORM when possible
3. skip some unnecessary validation checks (specifically when adding groundtruth data, we do not need to both check if the datum does not exist and then check if annotations do not exist, since datum creation and groundtruth annotation creation must happen together).
4. Add two indexes on the `Annotation` table, one for `model_id` and one for `datum_id`.

This optimization was spurred by seeing slow performance on the demo instance. I tested by running the API locally but pointing to the postgres hosted on the demo instance. For the example data I was using running off the `main` branch gave the following speeds:

- groundtruth creation: 517s
- prediction creation: 531s

with the code changes except for the additional index (i.e. just 1-3 above) the times became

- groundtruth creation: 12s
- prediction creation: 40s

adding those two indexes on top of this, the times became

- groundtruth creation: 12s
- prediction creation: 25s


api/valor_api/backend/core/datum.py
```diff
@@ -7,6 +7,29 @@
 from valor_api.backend.query import Query
 
 
+def create_datums(
+    db: Session, datums: list[schemas.Datum], datasets: list[models.Dataset]
+) -> list[models.Datum]:
```
ntlind, line 12: should we add a docstring here (similar to `create_datum`)?

api/valor_api/backend/core/groundtruth.py
```diff
@@ -6,70 +6,62 @@
...
-        != enums.TableStatus.CREATING
-    ):
-        raise exceptions.DatasetFinalizedError(groundtruth.dataset_name)
-
-    # create datum
-    datum = core.create_datum(db=db, datum=groundtruth.datum, dataset=dataset)
-
-    # create labels
+def create_groundtruths(db: Session, groundtruths: list[schemas.GroundTruth]):
+    # check dataset statuses
```
ntlind, line 10: should we add a docsting here?

ntlind (approved):

Issue #564, opened 2024-04-24
ENH: Better handling of bulk adds


ekorman:
### Feature Type

- [ ] Adding new functionality to valor

- [X] Changing existing functionality in valor

- [ ] Removing existing functionality in valor

### Problem Description

When bulk adding groundtruths or predictions, if there's some error (e.g. a network error in the post of one of the chunks or one of the groundtruths is bad), then the backend can get in a state where some of the requested data was added but not others. This makes it difficult for the user to know what data to retry.

### Feature Description

-

rsbowman-striveworks:
If a batch of datums is sent to Valor some of which have been previously added, the whole `add_groundtruths` call returns 409 'Datum already exists' and no new datums are added (tested with 0.21.2-post1).  This makes it difficult to ensure that all Chariot datums are added to the corresponding Valor dataset by using only the bulk interface.

In Chariot we have been using the one-at-a-time interface and ignoring the 'Datum already exists' errors in order to ensure that all datums are added to a possibly incomplete Valor dataset.  (Datasets can be incomplete because of errors and also from canceled bulk inference operations or training runs.)  It would be nice if we could have similar sematics for the bulk operations.

PR #580, opened 2024-05-14
add ignore_existing_datums argument for bulk uploads

ekorman:
This PR addresses #564 by

1. Allowing the user to pass an option to ignore existing datums when adding groundtruth
2. Get rid of the chunking strategy (put this on the onus of the user) since it can end up in a confusing state where only some groundtruths were added if there was an error in one of the chunks.

### DIFF
--- a/api/tests/functional-tests/backend/core/test_annotation.py
+++ b/api/tests/functional-tests/backend/core/test_annotation.py
@@ -89,7 +89,7 @@ def test_create_annotation_already_exists_error(
 
     core.create_groundtruths(db, empty_groundtruths)
     core.create_predictions(db, empty_predictions)
-    with pytest.raises(exceptions.DatumsAlreadyExistsError):
+    with pytest.raises(exceptions.DatumsAlreadyExistError):
         core.create_groundtruths(db, empty_groundtruths[0:1])
     with pytest.raises(exceptions.AnnotationAlreadyExistsError):
         core.create_predictions(db, empty_predictions[0:1])
--- a/api/tests/functional-tests/backend/core/test_annotation.py
+++ b/api/tests/functional-tests/backend/core/test_annotation.py
@@ -89,7 +89,7 @@ def test_create_annotation_already_exists_error(
 
     core.create_groundtruths(db, empty_groundtruths)
     core.create_predictions(db, empty_predictions)
-    with pytest.raises(exceptions.DatumAlreadyExistsError):
+    with pytest.raises(exceptions.DatumsAlreadyExistsError):
         core.create_groundtruths(db, empty_groundtruths[0:1])
     with pytest.raises(exceptions.AnnotationAlreadyExistsError):
         core.create_predictions(db, empty_predictions[0:1])
--- a/api/valor_api/backend/core/datum.py
+++ b/api/valor_api/backend/core/datum.py
@@ -44,33 +44,36 @@ def create_datums(
     ]
 
     try:
-        if ignore_existing_datums:
-            values = [
-                {
-                    "uid": row.uid,
-                    "dataset_id": row.dataset_id,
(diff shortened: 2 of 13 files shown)


api/valor_api/backend/core/datum.py
```diff
@@ -8,7 +8,10 @@
 
 
 def create_datums(
-    db: Session, datums: list[schemas.Datum], datasets: list[models.Dataset]
+    db: Session,
+    datums: list[schemas.Datum],
+    datasets: list[models.Dataset],
+    ignore_existing_datums: bool,
```
ntlind, line 14: nit: add this arg to the docstring?

api/tests/functional-tests/backend/core/test_annotation.py
```diff
@@ -89,7 +89,7 @@ def test_create_annotation_already_exists_error(
 
     core.create_groundtruths(db, empty_groundtruths)
     core.create_predictions(db, empty_predictions)
-    with pytest.raises(exceptions.DatumAlreadyExistsError):
+    with pytest.raises(exceptions.DatumsAlreadyExistsError):
```
ntlind, line 92: nit: `DatumsAlreadyExistError` is probably the more grammatically-correct verbage (removing the 's' from "Exists")

ntlind (approved):

api/valor_api/backend/core/datum.py
```diff
@@ -38,10 +41,39 @@ def create_datums(
...
+        for datum, dataset in zip(datums, datasets):
+            try:
+                existing_datums.append(fetch_datum(db, dataset.id, datum.uid))
+            except exceptions.DatumDoesNotExistError:
+                pass
+
+        if not ignore_existing_datums:
+            raise exceptions.DatumsAlreadyExistsError(
+                [datum.uid for datum in existing_datums]
+            )
```
czaloom, line 60: Instead of adding this flag could we instead cache all the datum uids that fail and then raise the exception after all the datums have been added?

This would remove the need for the extra param and still ensure that all new datums are ingested

czaloom (changes requested):

api/valor_api/backend/core/datum.py
```diff
@@ -36,12 +44,58 @@ def create_datums(
...
+                insert(models.Datum)
+                .values(values)
+                .on_conflict_do_nothing(index_elements=["dataset_id", "uid"])
+            )
+
+            db.execute(insert_stmt)
+
+            # get all the ids and update the row objects
+            for row in rows:
+                row.id = (
```
czaloom, line 66: Instead of iterating through all the rows again to retrieve ids you could try calling `.returning` on the insert statement.

See [reference](https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#getting-new-objects-with-returning).

```
insert_stmt = (
    insert(models.Datum)
    .values(values)
    .on_conflict_do_nothing(index_elements=["dataset_id", "uid"])
    .returning(models.Datum)
)
rows = db.scalars(insert_stmt).all()
```

ekorman: i tried that but it will only return the new ones and we need all of them

czaloom: Im not sure whether this will miss the conflicted datums

ekorman: oh wait i guess actually we only want to add annotations to the newly created ones.....

czaloom: I think youre right b/c the create annotations uses a zip which requires the ordered list...

```
for annotations_per_datum, datum, model in zip(
            annotations, datums, models_
        )
```

This should never be run if there exist duplicates

czaloom (approved):

PR #581, opened 2024-05-14
Add support for new filtering ops

czaloom:
# Feature

This PR implements a new filtering schema for Valor and connects the refactored query generator #597 to the API.

This includes support for logical operations (AND, OR, NOT) and the ability to apply filters per table.

# API Endpoints

Filters no longer use query parameters in api calls.

```
GET - gets all instances of a table 
/datasets
/models
/data
/labels

POST - post a filter, return filtered instances of a table
/datasets/filter
/models/filter
/data/filter
/labels/filter
```

# Examples

```
# No swimmer, small boats
no_swimmer_small_boats = client.get_datums(
    Filter(
        datums=And(
            Label.key == "class",
            Label.value != "swimmer",
        ),
        annotations=And(
            Label.key == "class",
            Label.value == "boat",
            Annotation.bounding_box.area < 50,
        ),
    )
)
assert len(no_swimmer_small_boats) == 1
assert no_swimmer_small_boats[0].uid == "uid1"

# No swimmer, large boats
no_swimmer_large_boats = client.get_datums(
    Filter(
        datums=And(
            Label.key == "class",
            Label.value != "swimmer",
        ),
        annotations=And(
            Label.key == "class",
            Label.value == "boat",
            Annotation.bounding_box.area > 50,
        ),
    )
)
assert len(no_swimmer_large_boats) == 1
assert no_swimmer_large_boats[0].uid == "uid2"

# Swimmer with small boat
swimmer_with_small_boats = client.get_datums(
    Filter(
        datums=And(
            Label.key == "class",
            Label.value == "swimmer",
        ),
        annotations=And(
            Label.key == "class",
            Label.value == "boat",
            Annotation.bounding_box.area < 50,
        ),
    )
)
assert len(swimmer_with_small_boats) == 1
assert swimmer_with_small_boats[0].uid == "uid3"

# Swimmer with large boat
swimmers_and_boats = client.get_datums(
    Filter(
        datums=And(
            Label.key == "class",
            Label.value == "swimmer",
        ),
        annotations=And(
            Label.key == "class",
            Label.value == "boat",
            Annotation.bounding_box.area > 50,
        ),
    )
)
assert len(swimmers_and_boats) == 1
assert swimmers_and_boats[0].uid == "uid4"
```


api/valor_api/schemas/__init__.py
```diff
@@ -5,14 +5,17 @@
...
-    StringFilter,
-    convert_filter_query_params_to_filter_obj,
+    FilterOperator,
+    LogicalFunction,
+    LogicalOperator,
+    SupportedSymbol,
+    SupportedType,
+    Symbol,
+    Value,
+    soft_and,
```
ntlind, line 17: what are "soft_and" and "soft_or"? could these names be more descriptive?

api/valor_api/main.py
```diff
@@ -233,13 +240,17 @@ def get_prediction(
...
 )
 def get_labels(
     response: Response,
-    filters: schemas.FilterQueryParams = Depends(),
-    offset: int = 0,
-    limit: int = -1,
+    offset: int = Query(
+        0, description="The start index of the items to return."
+    ),
+    limit: int = Query(
```
ntlind, line 250: is there a precedent for using this design pattern? saying limit is an int, but then defaulting it to a `Query` object seems very unnatural to me. I'm hoping this is a common pattern you can point me to in other repos

client/valor/client.py
```diff
@@ -404,7 +403,7 @@ def get_labels(
         """
         Gets all labels using an optional filter.
 
-        `GET` endpoint.
```
ntlind, line 407: remind me: why is it necessary to remove filter query params, which in turns makes this `GET` request into a `POST` request?

client/valor/schemas/symbolic/operators.py
```diff
@@ -1,223 +1,234 @@
...
 
-class Function:
-    """Base class for defining a function."""
+class Condition:
+    """Base class for defined a filter condition."""
 
-    _operator = None
+    def __init__(self, lhs: Any, rhs: Optional[Any] = None) -> None:
+        """
+        TODO
```
ntlind, line 10: fix TODOs everywhere in the repo? there were some outstanding ones from prior filtering PRs as well

client/valor/schemas/symbolic/operators.py
```diff
@@ -1,223 +1,234 @@
...
+        if isinstance(self._args, list):
+            raise ValueError("Negation only takes one argument.")
+        return self._args
 
 
-class IsNull(OneArgumentFunction):
+class IsNull(Condition):
     """Implementation of is null value check."""
 
     pass
```
ntlind, line 148: should these functions all just pass? I'm confused why we need to define these separately

integration_tests/client/datatype/test_data_generation.py
```diff
@@ -313,7 +312,7 @@ def test_generate_segmentation_data(
 
     for image in dataset.get_datums():
         uid = image.uid
-        sample_gt = dataset.get_groundtruth(uid)
+        sample_gt = dataset.get_groundtruth(uid)  # type: ignore - issue #604
```
ntlind, line 315: what's causing there to be a type issue on this line with these changes?

integration_tests/client/filtering/test_filtering_examples.py
```diff
@@ -0,0 +1,321 @@
...
+                Label.value == "boat",
+                Annotation.bounding_box.area > 50,
+            ),
+        )
+    )
+    assert len(swimmers_and_boats) == 1
+    assert swimmers_and_boats[0].uid == "uid4"
+
+
+# NOTE The example below does not work.
```
ntlind, line 205: commented code?

integration_tests/client/metrics/test_segmentation.py
```diff
@@ -107,9 +116,7 @@ def test_evaluate_segmentation_with_filter(
     color = Datum.metadata["color"]
     eval_job = model.evaluate_segmentation(
         dataset,
-        filter_by=[
-            color == "red",
-        ],
+        filters=Filter(datums=(color == "red")),  # type: ignore #605
```
ntlind, line 119: does this need to introduce a new type error?

api/tests/unit-tests/schemas/test_filter_schemas.py
```diff
@@ -1,159 +1,62 @@
...
-    ) == LessThan(lt=operands)
+    # if list has length of 1, return the contents
+    assert soft_or([condition]) == condition
 
-    assert DateTimeFilter(value=value, operator="<=",).to_function(
-        name
-    ) == LessThanEqual(le=operands)
-
-
-def test_geospatial_filter_to_function():
```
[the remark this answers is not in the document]
czaloom: These tests were for code that no longer exists! They translated the old filters format into  the new format for the query generator.

api/valor_api/main.py
czaloom: This is a fastapi defintion to specify that an arg is a query parameters. [Link](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#add-query-to-annotated-in-the-q-parameter)

client/valor/client.py
czaloom: Yes. B/c the new filters can sometimes be larger than is allowed in query params.

integration_tests/client/metrics/test_segmentation.py
czaloom: typo, this should be `issue #605`

api/tests/functional-tests/backend/core/test_label.py
```diff
@@ -346,18 +358,78 @@ def test_get_joint_labels(
     labels = get_joint_labels(
         db=db,
         lhs=schemas.Filter(
-            dataset_names=[dataset_name],
-            task_types=[enums.TaskType.CLASSIFICATION],
-            require_bounding_box=False,
-            require_polygon=False,
-            require_raster=False,
+            labels=schemas.LogicalFunction(
```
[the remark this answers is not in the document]
czaloom: I couldnt think of a more intuitive name. I could enumerate it into `And`, `Or` and `Not` like the client?

api/valor_api/main.py
ntlind: gotcha, thank you! I was confused because we used to have an object named `Query` too, but I think you got rid of it recently

api/tests/unit-tests/schemas/test_filter_schemas.py
```diff
@@ -1,159 +1,62 @@
...
 
-    assert NumericFilter(value=value, operator=">=",).to_function(
-        name
-    ) == GreaterThanEqual(ge=operands)
+def test_soft_and(condition: Condition):
 
-    assert NumericFilter(value=value, operator="<",).to_function(
-        name
-    ) == LessThan(lt=operands)
+    # raises value error if list is empty
```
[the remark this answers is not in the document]
czaloom: Im confused, this is a comment describing the next test

```
# raises value error if list is empty
        soft_and([None, None, None])
```

api/valor_api/schemas/__init__.py
czaloom: These functions handle the AND'ing of lists that were built using list comprehension and may not be valid.

I can rename them to `And` and `Or` since they basically are used as such.

czaloom: A better name could be `safe_and` and `safe_or`

client/valor/schemas/symbolic/operators.py
czaloom: Searched for and removed / completed todos.

client/valor/schemas/symbolic/operators.py
czaloom: These inherit `Condition` which initializes its op as `self.op = type(self).__name__.lower()`.

This allows client code to rely on a single source of truth (e.g. `Condition`)

integration_tests/client/datatype/test_data_generation.py
czaloom: Removed type ignore. There is red highlighting but it passes pre-commit.

api/valor_api/schemas/__init__.py
czaloom: I called it `soft_and` b/c it would only create an `And` if there was at least 2 elements. Otherwise it would pass the object through without modifying it.

api/tests/functional-tests/backend/core/test_label.py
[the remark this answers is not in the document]
ntlind: I just meant that we're going from one class `Filter` with a few easy arguments to 5(?) classes with more complicated arguments and doubling the lines of code in the process. I don't have any real suggestions here as I'm not deep enough in the code to understand if we need to be this specific with all of these different classes

api/tests/unit-tests/schemas/test_filter_schemas.py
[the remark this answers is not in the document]
ntlind: my bad, I misread these lines

integration_tests/client/datatype/test_data_generation.py
ntlind: I thought this was to be the last filtering PR, so I was expecting issues like this one to be fixed. I'm ok with pushing this into another PR.

integration_tests/client/filtering/test_filtering_examples.py
ntlind: why was this resolved? looks like there is still unnecessary commented code here

integration_tests/client/datatype/test_data_generation.py
czaloom: It is the last one to work on filter structure. These type ignores are related to pyright issues that only affect json generation in the client.

api/valor_api/schemas/__init__.py
ntlind: IMHO, I'd be as specific as possible (e.g., `and_if_contains_two_elements`). `safe_and`/`safe_or` could also be "good enough" IMO

api/tests/functional-tests/backend/core/test_label.py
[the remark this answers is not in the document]
czaloom: > 5(?) classes 

Which ones are you referring to?


`Filter`s are constructed using 2 classes. 
- `Condition` (e.g. ==, !=, >, etc) 
-  `LogicalFunction`s (e.g. AND, OR, NOT).

ekorman (approved):

integration_tests/client/filtering/test_geometric_filtering.py
```diff
@@ -151,13 +151,13 @@ def test_filter_by_bounding_box(client, groundtruths_with_areas, areas):
...
         thresholded_labels = client.get_labels(
-            Filter.create(
-                [
-                    Annotation.bounding_box.is_not_none(),
-                    Annotation.polygon.is_none(),
-                    Annotation.raster.is_none(),
-                    Annotation.bounding_box.area > area,
-                ]
+            Filter(
+                labels=(
```
ekorman, line 155: why are geometry filters being passed to `labels` argument?

api/valor_api/schemas/__init__.py
czaloom: I went in a different direction and implemented them as classmethods of `LogicalFunction`.

- `LogicalFunction.and_`
- `LogicalFunction.or_`
- `LogicalFunction.not_`

ekorman (approved):

Issue #582, opened 2024-05-16
ENH: Improve the scalability and specificity of PrecisionRecallCurve 


ntlind:
### Feature Type

- [ ] Adding new functionality to valor

- [X] Changing existing functionality in valor

- [ ] Removing existing functionality in valor

### Problem Description

Today, Valor's classification and object detection evaluations include a `PrecisionRecallCurve` metric if the user passes `compute_pr_curve=True` into their evaluation function. This output was designed to allow users to visualize example datums by hovering over the elements of a confusion matrix. The output of this metric typically looks like:

```python
{
    "type": "PrecisionRecallCurve",
    "parameters": {
        "label_key": "class", # The key of the label.
        "pr_curve_iou_threshold": 0.5, # Note that this value will be None for classification tasks. For detection tasks, we use 0.5 as the default threshold, but allow users to pass an optional `pr_curve_iou_threshold` parameter in their evaluation call.
    },
    "value": {
        "cat": { # The value of the label.
            "0.05": { # The confidence score threshold, ranging from 0.05 to 0.95 in increments of 0.05.
                "fn": [],
                "fp": [
                    (
                        'test_dataset',
                        1,
                        '{"type":"Polygon","coordinates":[[[464.08,105.09],[495.74,105.09],[495.74,146.99],[464.08,146.99],[464.08,105.09]]]}'
                    ) # There's one false positive for this (key, value, confidence threshold) combination as indicated by the one tuple shown here. This tuple contains that observation's dataset name, datum ID, and coordinates in the form of a GeoJSON string. For classification tasks, this tuple will only contain the given observation's dataset name and datum ID.
                ],
                "tp": [],
                "recall": -1,
                "f1_score": -1,
                "precision": 0.0,
            },
            ...
```

There are two issues with our current approach:
- *Scalability*: This output scales poorly as the number of datums increases since we're storing a unique bounding box in this deeply-nested map for each datum. We need to reduce the size of this map and/or chunk it up in such a way that downstream services don't have to consume the whole map at once.
- *Specificity*: Per [#576](https://github.com/Striveworks/valor/issues/576), it'd be helpful to know if a false positive / false negative was caused by:
    - a) a hallucination (we said there was an object, when there wasn't actually an object)
    - b) a misclassification (there was an object that we should have detected, but our prediction mislabeled it or gave it the wrong bounding box)
    - c) a missed detection (there was an object that we should have detected, but our model didn't output any predictions at all for a given datum)

### Feature Description

[...]

PR #584, opened 2024-05-21
Simplify `PrecisionRecallCurve` and add `DetailedPrecisionRecallCurve` for advanced debugging

ntlind:
# Improvements
- Updated `PrecisionRecallCurve` to only return counts for `fp`, `tp`, etc., rather than returning examples.
- Added `DetailedPrecisionRecallCurve` metric to implement the new functionality described in #582. Added `pr_curves_max_examples` to cap the number of datum examples returned to the user for scalability reasons (default=1).
- Added new docs describing how `misclassifications`, `hallucinations`, and `missed_detections` are calculated.

```python

pr_evaluation = evaluate_detection(
    data=dataset,
)
print(pr_evaluation)

[...,
{
    "type": "PrecisionRecallCurve",
    "parameters": {
        "label_key": "class", # The key of the label.
        "pr_curve_iou_threshold": 0.5, # Note that this value will be None for classification tasks. For detection tasks, we use 0.5 as the default threshold, but allow users to pass an optional `pr_curve_iou_threshold` parameter in their evaluation call.
    },
    "value": {
        "cat": { # The value of the label.
            "0.05": { # The confidence score threshold, ranging from 0.05 to 0.95 in increments of 0.05.
                "fn": 0,
                "fp": 1,
                "tp": 3,
                "recall": 1,
                "precision": 0.75,
                "f1_score": .857,
            },
            ...
        },
    }
},
{
    "type": "DetailedPrecisionRecallCurve",
    "parameters": {
        "label_key": "class", # The key of the label.
        "pr_curve_iou_threshold": 0.5,
    },
    "value": {
        "cat": { # The value of the label.
            "0.05": { # The confidence score threshold, ranging from 0.05 to 0.95 in increments of 0.05.
                "fp": {
                    "total": 1,
                    "observations": {
                        'hallucinations': {
                            "count": 1,
                            "examples": [
                                (
                                    'test_dataset',
                                     1,
                                    '{"type":"Polygon","coordinates":[[[464.08,105.09],[495.74,105.09],[495.74,146.99],[464.08,146.99],[464.08,105.91]]]}'
                               ) # There's one false positive for this (key, value, confidence threshold) combination as indicated by the one tuple shown here. This tuple contains that observation's dataset name, datum ID, and coordinates in the form of a GeoJSON string. For classification tasks, this tuple will only contain the given observation's dataset name and datum ID.
                            ],
                        }
                    },
                },
                "tp": {
                    "total": 3,
                    "observations": {
                        'all': {
                            "count": 3,
                            "examples": [
                                (
                                    'test_dataset',
                                     2,
                                    '{"type":"Polygon","coordinates":[[[464.08,105.09],[495.74,105.09],[495.74,146.99],[464.08,146.99],[464.08,105.91]]]}'
                               ) # We only return one example since `pr_curve_max_examples` is set to 1 by default; update this argument at evaluation time to store and retrieve an arbitrary number of examples.
                            ],
                        },
                    }
                },
                "fn": {...},
            },
        },
    }
}]
```


api/valor_api/schemas/metrics.py
```diff
@@ -418,7 +418,23 @@ def db_mapping(self, evaluation_id: int) -> dict:
...
+    Attributes
+    ----------
+    label_key: str
+        The label key associated with the metric.
+    pr_curve_iou_threshold: float, optional
+        The IOU threshold to use when calculating precision-recall curves. Defaults to 0.5.
+    """
+
+    label_key: str
+    pr_curve_iou_threshold: float | None = None
```
czaloom, line 434: This structure is still confusing me as it could easily conform to the standard metrics format.

```
class Metric(BaseModel):
    """
    A metric response from the API.

    Attributes
    ----------
    type : str
        The type of metric.
    parameters : dict
        The parameters of the metric.
    value : float
        The value of the metric.
    label : Label
        The `Label` for the metric.
    """

    type: str
    parameters: dict | None = None
    value: float | dict | None = None
    label: Label | None = None
```

But instead pulls label key up and stores a dictionary by label value?

@ekorman What are your thoughts? I remember there being a reason for this organization early on.

api/valor_api/backend/metrics/segmentation.py
```diff
@@ -202,27 +202,29 @@ def _compute_segmentation_metrics(
...
-            value=(
-                sum(iou_values) / len(iou_values)
-                if len(iou_values) != 0
-                else -1
-            ),
-            label_key=grouper_key,
-        )
-        for grouper_key, iou_values in ious_per_grouper_key.items()
-    ]
+    if "mIOU" in parameters.metrics:
```
czaloom, line 216: Are we not going to store the metrics that we computed? Is this desirable given the user could request these in the future? I feel that the computational cost of rerunning the evaluation to get a singular metric outweighs any memory cost of storing them even if they weren't requested in the moment.

I also feel that this stands for the detailed metrics. Why not add a `details` attribute to `schemas.Metric` and only populate the response with it if a flag is set?

api/valor_api/schemas/metrics.py
ntlind: responded via slack

api/valor_api/backend/metrics/segmentation.py
ntlind: responded via slack. storing intermediary metrics isn't advantageous since a second `evaluate_` request with new parameters would trigger a second evaluation.

ntlind: changed this PR to store all metrics per our discussion yesterday

api/valor_api/schemas/metrics.py
```diff
@@ -429,25 +445,76 @@ class PrecisionRecallCurve(BaseModel):
...
+    label_key: str
+        The label key associated with the metric.
+    value: dict
+        A nested dictionary where the first key is the class label, the second key is the confidence threshold (e.g., 0.05), the third key is the metric name (e.g., "precision"), and the final key is either the value itself (for precision, recall, etc.) or a list of tuples containing data for each observation.
+    pr_curve_iou_threshold: float, optional
+        The IOU threshold to use when calculating precision-recall curves. Defaults to 0.5.
+    """
+
+    value: dict[
+        str,  # '4'
```
czaloom, line 499: nit: I like the comments, but instead of `'4'` could you write something like `label value`. This goes for the other examples below too.

api/valor_api/backend/metrics/detection.py
```diff
@@ -61,6 +64,148 @@ def _calculate_101_pt_interp(precisions, recalls) -> float:
...
 
+def _calculate_ap_and_ar(
+    sorted_ranked_pairs: dict[str, list[RankedPair]],
+    number_of_groundtruths_per_grouper: dict[str, int],
+    grouper_mappings: dict[str, dict[str, schemas.Label]],
+    iou_thresholds: list[float],
+    recall_score_threshold: float,
+) -> Tuple[list[schemas.APMetric], list[schemas.ARMetric]]:
+    """
+    Computes the average precision. Return is a dict with keys
```
czaloom, line 75: nit: Update docstring to describe AR as well.

czaloom (approved):

PR #589, opened 2024-05-31
Remove `task_type` from `Annotation`; users should use `is_instance_segmentation=True` to denote instance segmentations 

ntlind:
# Improvements
- Remove `task_type` from `Annotations`
- Add `is_instance_segmentation` for use when leveraging rasters in object detection tasks 
- Remove client-side Prediction validations (e.g., checking that scores add up to 1 for classification annotations) since we already check for these errors on the API side


client/valor/coretypes.py
```diff
@@ -158,36 +156,6 @@ def __init__(
         """
         super().__init__(datum=datum, annotations=annotations)
 
-        # validation
-        for annotation in self.annotations:
```
ntlind, line 162: removed these checks as they seemed unnecessary given we already check for these things on the API side

api/tests/functional-tests/backend/core/test_label.py
```diff
@@ -584,68 +558,17 @@ def test_label_functions(
...
 
-    assert get_label_keys(
-        db,
-        schemas.Filter(
-            dataset_names=[dataset_name],
-            task_types=[enums.TaskType.OBJECT_DETECTION],
-        ),
-        ignore_predictions=True,
-    ) == {"inssegk1", "inssegk2", "inssegk3"}
-
```
czaloom, line 626: Im not super happy losing the ability to query by task type. 

With the upcoming filters pr we could implement `Annotation.task_type == TaskType.OBJECT_DETECTION` as...

```
Annotation.is_instance == True & (
    Annotation.bounding_box.isnot(None)
    | Annotation.polygon.isnot(None)
    | Annotation.raster.isnot(None)
)
```

czaloom: Maybe it would be a good idea to leave task type in the database. We could rename it to datatype to support the text annotations.

This would significantly reduce the size of the generated sql statements.

api/valor_api/backend/core/dataset.py
```diff
@@ -417,7 +404,6 @@ def get_dataset_summary(db: Session, name: str) -> schemas.DatasetSummary:
         ),
         num_polygons=get_n_groundtruth_polygons_in_dataset(db, name),
         num_rasters=get_n_groundtruth_rasters_in_dataset(db, name),
-        task_types=get_unique_task_types_in_dataset(db, name),
```
czaloom, line 420: I still feel like this is useful b/c it shows the type of data that stored.

api/valor_api/backend/core/geometry.py
```diff
@@ -84,7 +83,6 @@ def get_annotation_type(
             .join(models.Dataset, models.Dataset.id == models.Datum.dataset_id)
             .where(
                 models.Datum.dataset_id == dataset.id,
-                models.Annotation.task_type == task_type.value,
```
czaloom, line 87: This change is going to make it always returns raster if a semantic segmentation is stored with a bounding box detection. This will cause issues for the coco use case

czaloom: We have two options here
1. Enforce a singular data type per model
2. Store an inferred data type in the db to stratify over

api/tests/functional-tests/backend/core/test_label.py
ntlind: what's a practical example where we'd need to filter by task type? why can't we create a new filter (in a future PR) to check for `is_instance_segmentation=True`?

leaving `task_type` attached to `Annotations` via filters is confusing IMO

api/valor_api/backend/core/dataset.py
ntlind: data doesn't have "types" per this PR; the point was to separate the concern between datums and evaluation types. what's a practical example where this is useful?

api/valor_api/schemas/types.py
```diff
@@ -29,110 +22,90 @@
 ]
 
 
-def _check_if_empty_annotation(annotation: "Annotation") -> bool:
+def _match_annotation_to_implied_task_type(
```
czaloom, line 25: I feel that storing this implied data type would be very useful for filtering. It doesnt have to be exposed by the API (besides maybe DatasetSummary)

api/valor_api/backend/core/geometry.py
ntlind: > semantic segmentation is stored with a bounding box detection.

not sure what you mean here as semantic segmentation annotations can't be stored with a bounding box detection; it'll throw a value error in `_match_annotation_to_implied_task_type`

the coco notebook seemed to run fine. can you provide an example where this would fail?

api/valor_api/schemas/types.py
ntlind: I could get behind this idea, but let's save it for a separate PR that focuses on filtering enhancements

api/valor_api/backend/core/geometry.py
czaloom: The coco notebook currently splits semantic segmenation and object detection into their own models so this doesnt happen.

If you store them in the same model the return of this function is going to always be raster even for the obj det task.

Since the filtering has no way of seeing a difference the evaluation for obj detection will be run over the sem seg rasters.

api/valor_api/schemas/types.py
czaloom: I think too much functionality has been removed to approve, there needs to be a way to determine the data type in filters.

api/valor_api/backend/core/dataset.py
czaloom: In the ingestion validator you have an implied data type. Why not use this?

api/tests/functional-tests/backend/core/test_label.py
czaloom: From the standpoint of any of the getter function (`get_labels`, `get_datasets`, `get_models`, ...) this is problematic b/c you can no longer ask get me all the objects that have data for semantic segmenation or object detection.

ntlind: for discussion tomorrow: if I add `implied_task_types` to psql, can you add this filter back at your convenience?

integration_tests/client/metrics/test_detection.py
czaloom, line 1: This file is not specifying isinstance and yet it passes the test??? This is what is the cause of my concern b/c these are labeled `semantic_segmentations` but you run object detection eval over them

czaloom (approved):

PR #595, opened 2024-06-04
Add `metrics_to_return` to all evaluations

ntlind:
# Improvements
- Add `metrics_to_return` argument to let users specify what metrics to store, compute, and return in their evaluations. This parameter is optional, with the default being to return our standard, high-level metrics (e.g., Precision, Recall, etc.). We store all intermediate metrics, even if the user doesn't ask for them (e.g., we'll compute and store AP if the user only asks for mAP such that we can turn on caching later); we'll only return the metrics the user actually asked for in our evaluation. 
- Remove `compute_pr_curves` argument as users can now pass `PrecisionRecallCurve` to `metrics_to_return` instead.


api/valor_api/backend/core/evaluation.py
czaloom, line 1: I think we can implement the caching and reuse in one move here.

```
def _create_response(
    db: Session,
    evaluation: models.Evaluation,
    job_request: schemas.EvaluationRequest,
    **kwargs,
) -> schemas.EvaluationResponse:
    """Converts a evaluation row into a response schema."""
    metrics = db.query(
        select(models.Metric)
        .where(
            and_(
                models.Metric.evaluation_id == evaluation.id,
                models.Metric.type.in_(
                    job_request.parameters["metrics_to_return"]
                ),
            )
        )
```

api/valor_api/schemas/evaluation.py
```diff
@@ -50,6 +50,29 @@ class EvaluationParameters(BaseModel):
     def _validate_by_task_type(cls, values):
         """Validate the IOU thresholds."""
 
+        # set default metrics for each task type
+        if values.metrics_to_return is None:
+            match values.task_type:
```
czaloom, line 55: To allow caching the evaluation should be committed to the db with all metrics that will be computed. This allows the caching i talk about in my other comment.

PrecisonRecallCurve can be an edge case handled in its own statement here

czaloom (approved):

PR #598, opened 2024-06-04
prevent finalization over empty dataset

czaloom:
### Problem Description

Evaluations are currently allowed for empty datasets which stems from an empty dataset being valid for finalization.

### Feature Description

- Add a check for dataset finalization that ensures at least one datum exists.
  - Note: this already exists for model finalization.
- Change the behavior of evaluation retries as the logic is dependent on allowing empty datasets.


api/valor_api/backend/core/dataset.py
```diff
@@ -254,6 +254,15 @@ def set_dataset_status(
             dataset_names=[name],
         ):
             raise exceptions.EvaluationRunningError(dataset_name=name)
+    elif status == enums.TableStatus.FINALIZED:
+        datum_count = (
+            db.query(func.count(models.Datum.id))
```
ntlind, line 259: nit: this logic might be better encapsulated in a small, seperate function (e.g., `_validate_dataset_contains_datums`)

ntlind (approved):

PR #601, opened 2024-06-07
Remove sqlalchemy type ignores in the backend.

czaloom:
### Problem Description

The `Query` object is a major source of type ignores in the backend.

### Feature Description

Condense the `Query` generator object into two strongly typed functions: 
- `generate_select`
- `generate_query`

### Additional Context

**Before**
```
q = Query(
    models.Label.key,
    models.Label.value,
    models.Prediction.score,
).predictions(as_subquery=False)
q = q.where(models.Prediction.annotation_id == annotation.id)  # type: ignore - SQLAlchemy type issue
labels = [
    schemas.Label(
        key=scored_label[0],
        value=scored_label[1],
        score=scored_label[2],
    )
    for scored_label in db.query(q.subquery()).all()
]
```

**After**
```
query = (
    generate_query(            
        models.Label.key,
        models.Label.value,
        models.Prediction.score,
        db=db,
        label_source=models.Predication,
    )
    .where(models.Prediction.annotation_id == annotation.id)
)
labels = [
    schemas.Label(
        key=scored_label[0],
        value=scored_label[1],
        score=scored_label[2],
    )
    for scored_label in query.all()
]
```


api/valor_api/backend/core/dataset.py
```diff
@@ -165,9 +165,12 @@ def get_paginated_datasets(
             "Offset should be an int greater than or equal to zero. Limit should be an int greater than or equal to -1."
         )
 
-    datasets_subquery = (
-        Query(models.Dataset.id.label("id")).filter(filters).any()
-    )
+    advanced_filter = filters.to_advanced_filter() if filters else None
```
ntlind, line 168: the idea of an "advanced filter" is a bit confusing. can I assume this concept is still going away in a near-term PR?

api/valor_api/schemas/metrics.py
```diff
@@ -565,8 +592,10 @@ def db_mapping(self, evaluation_id: int) -> dict:
         ----------
         A mapping dictionary.
         """
+        value = self.value if self.value is not None else -1
```
ntlind, line 595: can we combine these two if statements so that we aren't assigning to `value` twice?

client/unit-tests/coretypes/test_filtering.py
```diff
@@ -76,7 +82,7 @@ def test__format_filter(geojson, polygon):
             Dataset.metadata["some_str"] == "foobar",
             Dataset.metadata["some_float"] >= 0.123,
             Dataset.metadata["some_datetime"] > datetime.timedelta(days=1),
-            Dataset.metadata["some_geospatial"].intersects(polygon),  # type: ignore
```
ntlind, line 79: you don't have to fix this schema issue in this PR, but can you add a ticket with context to the backlog so we don't forget about it?

client/unit-tests/symbolic/collections/test_structures.py
```diff
@@ -245,7 +245,7 @@ def test_list():
     }
 
     # test decode from json dict
-    assert List[Float].decode_value([0.1, 0.2, 0.3]).get_value() == [  # type: ignore
```
ntlind, line 248: same comment about adding ticket to GitHub

czaloom: https://github.com/Striveworks/valor/issues/604

client/unit-tests/coretypes/test_filtering.py
czaloom: #605

api/valor_api/backend/core/dataset.py
czaloom: `AdvancedFilter` will assume the name `Filter` when we make the switch at the API / Client level.

czaloom: I am considering it advanced as it is a superset of the existing `Filter`

ntlind (approved):

PR #608, opened 2024-06-07
Add `dataset_names` attribute to `EvaluationRequest`

czaloom:
### Feature Description

My proposed solution is to add `dataset_names: list[str]` as a parameter of `schemas.EvaluationRequest` and `schemas.EvaluationResponse` as well as adding a column of the same name to `models.Evaluation`.

The datatype of the `models.Evaluation` column will be of type `JSONB` and contain a list of strings.

This will enable the create / get endpoints for evaluations to take specific datasets and models as query params with the option to pass a filter in a json payload if desired. 

### Problem Description

After the switch to the new filtering schema, filters will no longer be transmitted via query params in the URL. The byproduct of this is that creating an evaluation for a simple model -> dataset(s) pairing will require the generation of a JSON payload.

With the complexity of the new filtering schema this does not seem like the best route as this is likely to be the most common method of interacting with evaluations. 

### Additional Context

This will affect all components of Valor.


api/tests/unit-tests/schemas/test_evaluation.py
```diff
@@ -102,14 +105,16 @@ def test_EvaluationRequest():
...
+            dataset_names=["ds"],
+            model_names=None,  # type: ignore - purposefully throwing error
+            filter=schemas.Filter(),
             parameters=schemas.EvaluationParameters(
                 task_type=enums.TaskType.CLASSIFICATION
             ),
         )
     with pytest.raises(ValidationError):
         schemas.EvaluationRequest(
+            dataset_names=["ds"],
```
ntlind, line 117: this test should probably check if dataset_names is `None` since `datum_filter` is no longer an attribute  of `EvaluationRequest`. can you ctrl + F for "datum_filter" in this file and make sure we aren't using it inappropriately on other tests?

client/valor/coretypes.py
```diff
@@ -172,6 +173,8 @@ def __init__(
         ----------
         id : int
             The ID of the evaluation.
+        dataset_names : list[str]
+            The names of the datasets the model was evaluated over.
         model_name : str
             The name of the evaluated model.
         datum_filter : schemas.Filter
```
ntlind, line 180: fix docstring; shouldn't include datum_filter, but should include `filter`

migrations/sql/00000009_add_dataset_names.up.sql
```diff
@@ -0,0 +1,13 @@
...
+DROP CONSTRAINT evaluation_model_name_datum_filter_parameters_key;
+
+UPDATE evaluation
+SET dataset_names = datum_filter->'dataset_names';
+
+UPDATE evaluation
+SET datum_filter = jsonb_set(datum_filter, '{dataset_names}', 'null'::jsonb);
+
+ALTER TABLE evaluation ADD CONSTRAINT evaluation_unique_constraint
+UNIQUE (model_name, datum_filter, parameters, dataset_names);
```
ntlind, line 13: should `datum_filter` be renamed to `filter`? does it only ever filter datums?

czaloom: `filter` is a keyword in SQL so we can't use it directly. Would `filters` work for you?

czaloom: This could also be applied to the schemas

ntlind: filters works for me, thanks!

ntlind (approved):

PR #611, opened 2024-06-11
have comparisons return bool when both objects are values

ekorman:
Currently, when two `Equatable` objects (that are not symbolic but have values) are compared, the result is a `Boolean` object instead of an actual `bool`. this causes some undesired behavior due to example `bool(Boolean(False)) == True`. and indeed `test_add_prediction` and `test_create_image_dataset_with_detections` were bad tests and actually should have been failing (see their updates)

This PR updates the following methods of `Equatable` to return bools when both sides have values:
- `__eq__`
- `__ne__`
- `in_`
- `__gt__`
- `__ge__`
- `__lt__`
- `__le__`


client/unit-tests/schemas/test_label.py
```diff
@@ -51,21 +51,21 @@ def test_scored_label():
...
-    assert (s1 == s2).get_value()  # type: ignore - resolved to 'Bool' as both sides are values
-    assert (s1 == s6).get_value()  # type: ignore - resolved to 'Bool' as both sides are values
-    assert not (s1 == s3).get_value()  # type: ignore - resolved to 'Bool' as both sides are values
-    assert not (s1 == s4).get_value()  # type: ignore - resolved to 'Bool' as both sides are values
-    assert not (s1 == s5).get_value()  # type: ignore - resolved to 'Bool' as both sides are values
+    assert s1 == s2  # type: ignore - resolved to 'Bool' as both sides are values
+    assert s1 == s6  # type: ignore - resolved to 'Bool' as both sides are values
+    assert not (s1 == s3)  # type: ignore - resolved to 'Bool' as both sides are values
+    assert not (s1 == s4)  # type: ignore - resolved to 'Bool' as both sides are values
+    assert not (s1 == s5)  # type: ignore - resolved to 'Bool' as both sides are values
```
czaloom, line 58: Can we remove all these type ignores now?

czaloom (approved):

client/unit-tests/schemas/test_label.py
ekorman: yep! removed

Issue #610, opened 2024-06-10
ENH: add client request timeout and possibly retry


rsbowman-striveworks:
### Feature Type

- [ ] Adding new functionality to valor

- [X] Changing existing functionality in valor

- [ ] Removing existing functionality in valor

### Problem Description

Valor client connections can hang.  

### Feature Description

Requests should use a [timeout](https://requests.readthedocs.io/en/stable/user/quickstart/#timeouts).  It's not important for us that the timeout is configurable, it can just be some reasonable value (10s?).  

Since we're adding a timeout, it might be worth adding retries with backoff, at least for idempotent requests.

ntlind:
From Scott: consider [swapping requests with urllib3](https://striveworks.slack.com/archives/C04JJ15357U/p1718118052953219?thread_ts=1718046030.825299&cid=C04JJ15357U)

ntlind:
From Scott: should we consider swapping `requests` with `urllib3`?

PR #615, opened 2024-06-11
Add retry logic to avoid `request` timeouts

ntlind:
# Improvements
- Address #610 by allowing `_requests_wrapper` to retry up to `max_retries_on_timeout` times if a request times out.

```
# max_retries_on_timeout, initial_timeout, and exponential_backoff are new arguments
def _requests_wrapper(
        self,
        method_name: str,
        endpoint: str,
        ignore_auth: bool = False,
        max_retries_on_timeout=2,
        initial_timeout: float = 2,
        exponential_backoff: int = 2,
        *args,
        **kwargs,
    ):
        """
        Wrapper for handling API requests.

        Parameters
        ----------
        method_name : str
            The name of the method to use for the request.
        endpoint : str
            The endpoint to send the request to.
        ignore_auth : bool, default=False
            Option to ignore authentication when you know the endpoint does not
            require a bearer token. this is used by the `_get_access_token_from_username_and_password`
            to avoid infinite recursion.
        max_retries_on_timeout: int
            The maximum number of retries when the requests module returns a Timeout error.
        initial_timeout : float
            The initial timeout value between how often we'll retry request methods.
        exponential_backoff : integer
            The factor by which we multiple initial_timeout for each subsequent retry.
        """
```

# Notes
I looked into swapping requests with urllib3 per Scott's suggestion on slack. I decided not to make this change because:
- requests is basically just urllib3 with syntactic sugar bundled and helpful summary features (e.g., `resp['detail']`). we do use some of these extra features in Valor, so we'd have to recreate them if we want to switch to urllib3
- requests is frequently updated (last commit was 2 weeks ago), is maintained by the Python Software Foundation, and has 50.1k stars on GH; seems fair to say that `requests` is production-ready

### DIFF
--- a/client/valor/client.py
+++ b/client/valor/client.py
@@ -193,6 +193,9 @@ class ClientConnection:
         method_name: str,
         endpoint: str,
         ignore_auth: bool = False,
+        max_retries_on_timeout=2,
+        initial_timeout: float = 2,
+        exponential_backoff: int = 2,
         *args,
         **kwargs,
     ):
@@ -209,6 +212,12 @@ class ClientConnection:
             Option to ignore authentication when you know the endpoint does not
             require a bearer token. this is used by the `_get_access_token_from_username_and_password`
             to avoid infinite recursion.
+        max_retries_on_timeout: int
+            The maximum number of retries when the requests module returns a Timeout error.
+        initial_timeout : float
+            The initial timeout value between how often we'll retry request methods.
+        exponential_backoff : integer
+            The factor by which we multiple initial_timeout for each subsequent retry.
         """
         accepted_methods = ["get", "post", "put", "delete"]
         if method_name not in accepted_methods:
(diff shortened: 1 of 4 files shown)


client/valor/client.py
```diff
@@ -224,13 +233,29 @@ def _requests_wrapper(
...
+            except requests.exceptions.Timeout as e:
+                if timeout_retries < max_retries_on_timeout:
+                    time.sleep(
+                        initial_timeout
+                        * exponential_backoff**timeout_retries
+                    )
+                    timeout_retries += 1
+                    continue
+                else:
+                    raise TimeoutError(str(e))
```
rsbowman-striveworks, line 257: What is the advantage of this over just `raise e`?

rsbowman-striveworks (approved):
Looks nice, one small question

client/valor/client.py
ntlind: it just passes the error message from `Timeout` (a specific exception from `requests.exceptions`) to `TimeoutError` (a generic built-in exception). not necessary, but likely contains additional details beyond the generic exception message

rsbowman-striveworks: Gotcha, makes sense :+1: .

czaloom (approved):

PR #618, opened 2024-06-12
Migrations Patch

czaloom:
Resolves migration issues related to the following:
- `EvaluationParameters`
  - removed `compute_pr_curve`
  - added `metrics_to_return`
  - added `pr_curve_iou_threshold`
- conversion for `PrecisionRecallCurve` metric


migrations/sql/00000011_pr_curve_metrics.up.sql
```diff
@@ -0,0 +1,41 @@
...
+RETURNS jsonb LANGUAGE plpgsql AS $$
+DECLARE
+    label_value text;
+    score_threshold text;
+    metric_key text;
+    metric_value jsonb;
+    label_value_dict jsonb;
+    score_threshold_dict jsonb;
+    metric_dict jsonb;
+BEGIN
```
ntlind, line 20: nit: could we use json path expressions instead of nested for loops?

ntlind (approved):

PR #633, opened 2024-06-25
Increase code coverage to 90%

ntlind:


api/valor_api/backend/core/llm_clients.py
```diff
@@ -163,15 +166,12 @@ def __call__(
                 messages=processed_messages,
                 seed=self.seed,
             )
-        # TODO Should we catch this if we aren't going to do anything?
-        except openai.BadRequestError as e:
+        except Exception as e:
```
ntlind, line 169: I made this more generic since BadRequestError is hard to test (requires extra inputs like `response` and `body`)

bnativi (approved):
lgtm

api/valor_api/backend/core/llm_clients.py
```diff
@@ -2,22 +2,28 @@
 
 import openai
 from mistralai.client import MistralClient
-from mistralai.exceptions import MistralAPIException
 from mistralai.models.chat_completion import ChatMessage
 
-COHERENCE_INSTRUCTION = """You are a helpful assistant. You will grade the user's text. Your task is to rate the text based on its coherence. Please make sure you read and understand these instructions carefully. Please keep this document open while reviewing, and refer to it as needed.
+
+def _get_instruction(metric: str):
```
bnativi, line 8: I like the  idea here of collecting the instructions into one function, however on the add_llm_guided_metrics branch, I added some documentation explaining the metric. I'm concerned that if we have details about every metric, then this function will look like a mess. Might be better to keep these as separate functions like _get_coherence_instruction() ?

PR #638, opened 2024-06-27
Add benchmarking workflow to evaluate classification and OD runtimes

ntlind:
## Use Cases
- We want to look at recent commits to figure out which one caused a change in performance
- We want to quickly test performance of potential changes locally using standardized benchmarks

## Changes
- Delete `stress-testing.yml` and associated files since we never trigger these manual jobs
- Create an automatic benchmarking job to track Valor performance for classification and object detection evaluations. The workflow will fail if any of the individual evaluation tests takes longer than 30 seconds.

## Output
image

image


integration_tests/benchmarks/object-detection/data.json
czaloom, line 1: Im not a fan of storing large data files in the codebase, is there a way to pull this from an external source?

.github/workflows/benchmark-evaluations.yml
```diff
@@ -0,0 +1,50 @@
...
+on:
+  push:
+    branches: "**"
+
+permissions:
+  id-token: write
+  contents: read
+
+env:
+  AWS_ROLE: arn:aws:iam::724664234782:role/Striveworks-Role-github_runner_npe
```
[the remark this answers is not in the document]
ntlind: this was copied directly from an existing workflow..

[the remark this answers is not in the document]
czaloom: This seems to be an oversight. The only workflow that had this was `stress-test.yml` which as you pointed out was unused.

czaloom (approved):

integration_tests/benchmarks/object-detection/data.json
czaloom: I think there are also reasonable concerns about this being a Github workflow.

In testing performance of v0.27.3 i currently have a run that is past the 1 hour mark. 

If a bug is introduced a new job could be spun up on each commit and cause multiple hours-long jobs.

ntlind: using how many datums? this shouldn't be a real issue if the limit is set low. 

eating up compute time is much less important then recognizing we have serious bugs that 60x our evaluation time

ekorman:
we can host the data in a public s3. prefer not having in version control

ekorman:
> we can host the data in a public s3. prefer not having in version control

e.g. i put this file https://github.com/Striveworks/valor/blob/bc73fb46ee52037980c8c146e4ad4ad9dc370b8b/integration_tests/benchmarks/object-detection/data.json

in our public bucket here:
https://pub-fa9e78b320c345228ff1a5785fbc07c8.r2.dev/data.json

ekorman (approved):

PR #642, opened 2024-07-01
Add Vacuum Analyze to Data Finalization

czaloom:
On dataset or model finalization a `VACUUM ANALYZE` is called to force the updating of database statistics.

This ensures that any subsequent evaluations will run with consistent completion times.

### DIFF
--- a/api/valor_api/backend/database.py
+++ b/api/valor_api/backend/database.py
@@ -26,6 +26,21 @@ logger.debug(
 engine = create_engine(SQLALCHEMY_DATABASE_URL)
 
 
+def vacuum_analyze():
+    pconn = None
+    pcur = None
+    try:
+        pconn = psycopg2.connect(SQLALCHEMY_DATABASE_URL)
+        pcur = pconn.cursor()
+        pconn.autocommit = True
+        pcur.execute("VACUUM ANALYZE;")
+    finally:
+        if pcur is not None:
+            pcur.close()
+        if pconn is not None:
+            pconn.close()
+
+
 def try_to_enable_gdal_drivers(db: Session) -> None:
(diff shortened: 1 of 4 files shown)


api/valor_api/backend/database.py
```diff
@@ -26,6 +26,25 @@
 engine = create_engine(SQLALCHEMY_DATABASE_URL)
 
 
+def vacuum_analyze():
```
ntlind, line 29: add docstring

api/valor_api/crud/_update.py
```diff
@@ -21,3 +29,8 @@ def finalize(*, db: Session, dataset_name: str, model_name: str | None = None):
             name=dataset_name,
             status=enums.TableStatus.FINALIZED,
         )
+
+    if task_handler:
+        task_handler.add_task(vacuum_analyze)
```
ntlind, line 34: what's the rationale for only doing this on finalization? would there be some benefit to doing it after x evaluation runs?

czaloom: The rationale for adding vacuum analyze to only the dataset/model finalization is based on the magnitude of table restructuring. The data added in dataset/model creation is used directly in computations so understanding the statistics of those tables (and related ones) is important to ensuring a consistent evaluation time.

Evaluations themselves do not produce data other than metrics. These metrics don't affect computation time and therefore there is little to no benefit to adding the same operation to them since we already implement indexing on these tables.

ntlind (approved):

api/valor_api/crud/_update.py
ntlind: makes sense! thanks for the write-up

PR #643, opened 2024-07-01
Fixed Bulk Add `DatumAlreadyExistsError` and `AnnotationAlreadyExistsError`

czaloom:
# Issue
The default timeout was 10s for all requests in the python client. For most requests this is fine, but for bulk add where a large JSON payload would have to be parsed by the API this would cause an early timeout.

This would have been caught earlier but the python client would perform automatic retrying of the POST request. This would immediately cause a conflict on the API-side as the already existing data was being uploaded.

### Example
The following shows the same request being resent without the knowledge of the user. The "retry" immediately fails because it directly conflicts with the original request.
```
{"method": "POST", "path": "/predictions", "hostname": "0.0.0.0", "status": 200, "duration_ms": 12469.926136996946, ...
{"method": "POST", "path": "/predictions", "hostname": "0.0.0.0", "status": 409, "duration_ms": 557.2792300081346, ...
```

# Solution
- remove retrying client requests, this exposes the issue so the user can retry the command themselves.
- expose timeout at the highest level so that it can be dynamically set by the user.

### DIFF
--- a/client/valor/client.py
+++ b/client/valor/client.py
@@ -189,6 +189,7 @@ class ClientConnection:
 
     def _requests_wrapper(
         self,
+        *_,
         method_name: str,
         endpoint: str,
         ignore_auth: bool = False,
@@ -222,6 +223,8 @@ class ClientConnection:
             raise ValueError(
                 f"method_name should be one of {accepted_methods}"
             )
+        elif method_name == "post" and max_retries != 0:
+            raise ValueError("POST requests cannot be automatically retried.")
 
         if endpoint[0] == "/":
             raise ValueError(
--- a/client/valor/client.py
+++ b/client/valor/client.py
@@ -195,7 +195,6 @@ class ClientConnection:
         timeout: Optional[float] = 2,
         max_retries=2,
         exponential_backoff: int = 2,
-        *args,
         **kwargs,
     ):
         """
@@ -245,7 +244,6 @@ class ClientConnection:
                     url,
                     headers=headers,
                     timeout=timeout,
-                    *args,
                     **kwargs,
                 )
             except requests.exceptions.Timeout as e:
@@ -281,7 +279,6 @@ class ClientConnection:
         endpoint: str,
         timeout: Optional[float] = 2,
         max_retries: int = 0,
-        *args,
         **kwargs,
     ):
         """
@@ -290,10 +287,9 @@ class ClientConnection:
         return self._requests_wrapper(
             method_name="post",
             endpoint=endpoint,
-            max_retries_on_timeout=max_retries,
-            initial_timeout=timeout,
-            *args,
-            *kwargs,
+            max_retries=max_retries,
+            timeout=timeout,
+            **kwargs,
         )
 
     def _requests_get_rel_host(self, endpoint: str, *args, **kwargs):
--- a/client/valor/client.py
+++ b/client/valor/client.py
@@ -242,7 +242,11 @@ class ClientConnection:
 
             try:
                 resp = requests_method(
-                    url, headers=headers, timeout=10, *args, **kwargs
+                    url,
+                    headers=headers,
+                    timeout=initial_timeout,
+                    *args,
+                    **kwargs,
                 )
             except requests.exceptions.Timeout as e:
                 if timeout_retries < max_retries_on_timeout:
@@ -273,12 +277,24 @@ class ClientConnection:
 
         return resp
 
-    def _requests_post_rel_host(self, endpoint: str, *args, **kwargs):
+    def _requests_post_rel_host(
+        self,
+        endpoint: str,
+        timeout: int = 30,
+        max_retries_on_timeout: int = 0,
+        *args,
+        **kwargs,
+    ):
         """
         Helper for handling POST requests.
         """
         return self._requests_wrapper(
-            method_name="post", endpoint=endpoint, *args, **kwargs
+            method_name="post",
+            endpoint=endpoint,
+            max_retries_on_timeout=max_retries_on_timeout,
+            initial_timeout=timeout,
+            *args,
+            *kwargs,
         )
 
     def _requests_get_rel_host(self, endpoint: str, *args, **kwargs):
--- a/client/valor/client.py
+++ b/client/valor/client.py
@@ -138,6 +138,11 @@ class ClientConnection:
         else:
             self._using_username_password = False
 
+        if self.access_token or self._using_username_password:
+            self._using_auth = True
+        else:
+            self._using_auth = False
+
         # check the connection by getting the api version number
         try:
             api_version = self.get_api_version()
@@ -193,9 +198,7 @@ class ClientConnection:
         method_name: str,
         endpoint: str,
         ignore_auth: bool = False,
-        timeout: Optional[float] = 2,
-        max_retries=2,
-        exponential_backoff: int = 2,
+        timeout: Optional[float] = 10,
         **kwargs,
     ):
         """
@@ -212,19 +215,13 @@ class ClientConnection:
             require a bearer token. this is used by the `_get_access_token_from_username_and_password`
             to avoid infinite recursion.
         timeout : float, optional
-            The initial timeout value between how often we'll retry request methods.
-        max_retries: int
-            The maximum number of retries when the requests module returns a Timeout error.
-        exponential_backoff : integer
-            The factor by which we multiple initial_timeout for each subsequent retry.
+            An optional request timeout in seconds.
         """
         accepted_methods = ["get", "post", "put", "delete"]
         if method_name not in accepted_methods:
             raise ValueError(
                 f"method_name should be one of {accepted_methods}"
             )
-        elif method_name == "post" and max_retries != 0:
-            raise ValueError("POST requests cannot be automatically retried.")
 
         if endpoint[0] == "/":
             raise ValueError(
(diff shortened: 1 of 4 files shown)


client/valor/client.py
```diff
@@ -192,10 +192,9 @@ def _requests_wrapper(
         method_name: str,
         endpoint: str,
         ignore_auth: bool = False,
-        max_retries_on_timeout=2,
-        initial_timeout: float = 2,
+        timeout: Optional[float] = 2,
```
ntlind, line 195: move optional parameter to the bottom of the parameter list?

client/valor/coretypes.py
```diff
@@ -1458,6 +1471,7 @@ def create_predictions(
         dataset: Dataset,
         model: Model,
         predictions: List[Prediction],
+        timeout: Optional[float] = None,
```
ntlind, line 1474: which functions do you expect to have retry logic after this PR, and which functions won't have retry logic? it's a little hard to follow here since the default is two seconds in `_requests_wrapper` but you're overriding higher-level defaults to None.

will this new logic sufficiently address #610?

czaloom: >  since the default is two seconds in _requests_post_rel_host

The default is 2s for generic POST requests.

>  but you're overriding higher-level defaults to None.

The default for creating predictions and groundtruths is None and is also user configurable. This is because it should scale with the number of objects included in the bulk create.

> which functions do you expect to have retry logic after this PR

The source of the bug this PR is addressing is automatic POST request retrying. I have removed the ability to auto-retry POSTing so that the user can themselves attempt to retry.

czaloom: > will this new logic sufficiently address https://github.com/Striveworks/valor/issues/610?

Yes it should, Ill update the create default to 10s by default since that was specified.

client/valor/client.py
czaloom: This is pretty nitty. 

Could you reference a standard that says this?

A value of None disables the timeout and disables the parameters below it so this seems to be the more intuitive ordering.

ntlind: this is a general requirement for many other programming languages (e.g., JS, but not Python) that take positional arguments. if the user passes positional arguments instead of kw arguments and your `timeout` isn't at the end, then you're essentially making `timeout` mandatory instead of optional.

czaloom: But none of the other arguments are positional... they are all kw args.

client/valor/coretypes.py
ntlind: > The source of the bug this PR is addressing is automatic POST request retrying. I have removed the ability to auto-retry POSTing so that the user can themselves attempt to retry.

does making it user-configurable give users a footgun (i.e., will they likely run into bugs if they set `timeout` manually)? I'm wondering if it makes more sense to remove the retry logic entirely, rather than turn it on for some functions and make it configurable for others

client/valor/client.py
ntlind: > But none of the other arguments are positional... they are all kw args

all python functions allow for positional arguments, we just mostly use kwargs on the backend. imagine a future Valor dev who isn't as careful but only uses positional args (and not any kwargs): make their life easier by making all of these args optional, instead of just `timeout`

client/valor/coretypes.py
czaloom: I didnt make the retry logic configurable, retrying for POST requests is completely removed. The timeout is what is now user configurable for large bulk requests.

This makes sense because what is "bulk" could be 10 datums or 10,000 datums.

client/valor/client.py
ntlind: this is one of the reasons you'll people use `-1` to mean "don't use a timeout" instead of `None`

czaloom: The issue of people new to the codebase mistakenly setting kw parameters in the client backend seems like a bit of a reach. 

I just pushed a commit that disables vague `args` for all the requests wrappers in the `client.py` file as they are all somewhat vague anyway. 

The GET, POST, PUT and DELETE wrappers now take endpoint as a positional and all other params must be kwargs.

czaloom: > this is one of the reasons you'll see people use 0 or -1 to mean "don't use a timeout" instead of None

I agree, but the requests library accepts `float | None` for its timeout kwarg.

client/valor/coretypes.py
ntlind: > I didnt make the retry logic configurable, retrying for POST requests is completely removed.

take `create_groundtruths`: how is retrying for POST requests completely removed if the user can configure them with a `timeout` param?

czaloom: The user can configure the timeout for the singular request that is sent when calling `add_groundtruths`. 

This has nothing to do with retrying. If the timeout occurs or another exception is raised, the user can make the best decision for themselves on how to move forward.

Furthermore, with how variable postgres ingestion seems an ingestion of 10 datums can take less than a second or 30 seconds so there is no way to pre-bake an effective timeout. It it 100% dependent on the user's hardware.

client/valor/client.py
ntlind: sure, but requests also allows you to pass `0` to mean "no timeout". this would be way less confusing then the way it's written now, right? then you just have to check if `timeout > 0`

> See requests src they even put a non-optional kwarg after...

"non-optional" means "has no default"; all the args after `timeout` in `requests` have a default, so this statement isn't true.

all of your args after `timeout` here all have a defaults, so (I believe) they are also all technically optional. you're specifying `timeout` as `Optional`, but not the others, which is confusing: might as well say they're all `Optional`

czaloom: Would changing it to Union[float, None] work better for you? Because the value is not optional, it has a default of 2 and can be set to None to disable it.

czaloom: Also i cant find it in the docs that setting it to zero disables it.

czaloom: > all of your args after timeout here all have a defaults, so (I believe) they are also all technically optional. you're specifying timeout as Optional, but not the others, which is confusing: might as well say they're all Optional

Yes but in python `typing.Optional[float]` is an alias of `typing.Union[float, None]` so it doesnt make sense at all to tag those optional b/c `None` is NOT a valid value for them. If the user doesnt specify them they use their defaults.

czaloom: Also we are including generic kwargs so the optional isnt even the last kwarg...

but if changing 
```
        ignore_auth: bool = False,
        timeout: Optional[float] = 2,
        max_retries=2,
        exponential_backoff: int = 2,
        **kwargs,
```
to
```
        ignore_auth: bool = False,
        max_retries=2,
        exponential_backoff: int = 2,
        timeout: Optional[float] = 2,
        **kwargs,
```
unblocks this for you then ill make the change

ntlind: > Would changing it to Union[float, None] work better for you? Because the value is not optional, it has a default of 2 and can be set to None to disable it.

that would be less confusing, thanks. the best strategy here IMO is just not to use `None` to mean "no timeout", but you seem pretty set on `None`.

[here is a prior issue](https://github.com/psf/requests/issues/1615) from 2013 which states that `timeout=0` means "no timeout".

client/valor/coretypes.py
ntlind: I'm not following what you mean here, and you didn't seem to engage with my `create_groundtruths` example; maybe we chat about it tomorrow?

integration_tests/client/test_client.py
```diff
@@ -205,8 +205,8 @@ def _return_mock_response(*args, **kwargs):
             method_name="get",
             endpoint="test",
             ignore_auth=False,
-            max_retries_on_timeout=max_retries,
-            initial_timeout=0.1,
```
ntlind, line 209: can you add a test to check your "no timeout" functionality (whether you use `None` or `0`)?

client/valor/client.py
czaloom: That issue contains the following 

>  they are thoroughly distinct values in this context. I would be okay with seeing timeout=0 and immediately raising a Timeout exception before doing anything. Since we don't return a response (or request) this won't affect the API.

client/valor/coretypes.py
czaloom: `Dataset.add_groundtruths` calls `client.add_groundtruths` so my comment above directly addresses it

> The user can configure the timeout for the singular request that is sent when calling add_groundtruths.

This has nothing to do with retrying. If the timeout occurs or another exception is raised, the user can make the best decision for themselves on how to move forward.

client/valor/client.py
ntlind: if you read the bottom of the issue, it says this was fixed and merged back in 2013. the issue is helpful because it communicates the devs intent.

czaloom: There was no code associated with that issue. The linked urllib3 pr at the end also shows `timeout=0` raises a `ReadTimeoutError`.

czaloom: Also should of thought of this earlier but you can literally see it doesnt work by trying it

```
>>> requests.get("https://google.com")
<Response [200]>
>>> requests.get("https://google.com", timeout=0)
ValueError: Attempted to set connect timeout to 0, but the timeout cannot be set to a value less than or equal to 0.
```

integration_tests/client/test_client.py
czaloom: Setting `timeout=0` raises a value error when using the requests library.

```
>>> requests.get("https://google.com")
<Response [200]>
>>> requests.get("https://google.com", timeout=0)
ValueError: Attempted to set connect timeout to 0, but the timeout cannot be set to a value less than or equal to 0.
```

czaloom: I added two tests!

client/valor/client.py
ntlind: good catch. you could always check for 0 or -1 instead of checking for `None` to avoid this whole issue. you seem pretty set on using `None`, and I'm not that invested in this change

client/valor/coretypes.py
ntlind: I see you're now throwing a ValueError to catch any POST retries; thanks for adding that change!

I now see what you mean by some of your comments above; I was conflating the timeout logic vs. the retry logic. I'm good with these changes; thanks for adding additional tests

ntlind (approved):

client/valor/client.py
```diff
@@ -232,81 +231,124 @@ def _requests_wrapper(
...
+            The endpoint to send the request to.
+        json : dict | list, optional
+            An optional kwarg to pass to the request.
+        params : dict, optional
+            An optional kwarg to pass to the request.
+        data : dict, optional
+            An optional kwarg to pass to the request.
+        ignore_auth : bool, default=False
+            Option to ignore authentication when you know the endpoint does not
+            require a bearer token. this is used by the `_get_access_token_from_username_and_password`
```
ntlind, line 278:
```suggestion
            require a bearer token. This is used by the `_get_access_token_from_username_and_password`
```

ntlind (approved):

ekorman (approved):

ekorman (approved):

PR #645, opened 2024-07-01
Fix `validate_matching_label_keys`

czaloom:
`validate_matching_label_keys` was accessing data that was out of scope. This PR implements a hard boundary on the datasets and model involved in an evaluation.

### DIFF
--- a/api/valor_api/backend/core/label.py
+++ b/api/valor_api/backend/core/label.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm import InstrumentedAttribute, Query, Session
 
 from valor_api import api_utils, schemas
 from valor_api.backend import models
-from valor_api.backend.query import generate_query, generate_select
+from valor_api.backend.query import generate_query
 from valor_api.backend.query.types import TableTypeAlias
 
 LabelMapType = list[list[list[str]]]
@@ -15,9 +15,9 @@ LabelMapType = list[list[list[str]]]
 
 def validate_matching_label_keys(
     db: Session,
+    dataset_names: list[str],
+    model_name: str,
     label_map: LabelMapType | None,
-    prediction_filter: schemas.Filter,
-    groundtruth_filter: schemas.Filter,
 ) -> None:
     """
     Validates that every datum has the same set of label keys for both ground truths and predictions. This check is only needed for classification tasks.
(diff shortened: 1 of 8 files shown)


api/valor_api/backend/core/label.py
```diff
@@ -26,71 +26,95 @@ def validate_matching_label_keys(
...
             ),
         )
-        .select_from(gts)
-        .group_by(gts.c.datum_id)
+        .select_from(models.Datum)
+        .join(
+            models.Dataset,
+            and_(
+                models.Dataset.id == models.Datum.dataset_id,
+                models.Dataset.name.in_(dataset_names),
```
ntlind, line 54: could the filters be an important addition here? is there an edge case where selecting all labels from a dataset/model wouldn't pass this validation, but filtering the labels using some user's filters would pass this validation?

api/valor_api/backend/core/label.py
```diff
@@ -26,71 +26,95 @@ def validate_matching_label_keys(
...
         Optional mapping of individual labels to a grouper label. Useful when you need to evaluate performance using labels that differ across datasets and models.
 
-
     Raises
     -------
     ValueError
         If the distinct ground truth label keys don't match the distinct prediction label keys for any datum.
     """
 
-    gts = generate_select(
```
ntlind, line 43: we use `generate_select` statements like this one in a few other places; could those be leaking data, too?

czaloom: Im reverting to using filters per your previous comment. These should not leak as it is now adding datasets, model, label_mapping and task_type to the select generation.

These were missing previously.

api/valor_api/backend/core/label.py
czaloom: Ill revert to using `from valor_api.backend.metrics.metric_utils import prepare_filter_for_evaluation` as it properly formats the filter for evaluation.

client/valor/client.py
```diff
@@ -273,36 +273,36 @@ def _requests_wrapper(
...
         """
         Helper for handling POST requests.
         """
         return self._requests_wrapper(
-            method_name="post", endpoint=endpoint, *args, **kwargs
+            method_name="post", endpoint=endpoint, **kwargs
         )
 
-    def _requests_get_rel_host(self, endpoint: str, *args, **kwargs):
+    def _requests_get_rel_host(self, endpoint: str, *_, **kwargs):
```
ntlind, line 284: can you explain these "*args" -> "*_" changes a bit more? is this related to `validate_matching_label_keys` somehow?

api/valor_api/backend/core/label.py
ntlind: thank you!

api/valor_api/backend/core/label.py
ntlind: great, thanks!

client/valor/client.py
ntlind: it seems like your client and integration changes are unrelated to this PR, and may be leaked here from another PR. just want to confirm that's true. best practice would be to remove them, but I'd consider it a nit

czaloom: Removed! They got dragged in by an unrelated pr.

ntlind (approved):

PR #651, opened 2024-07-03
Implement `valor_core` to compute metrics locally via numpy

ntlind:
## Improvements
- Create a local package, `valor_core`, which can evaluate classification and object detection tasks locally in numpy without using postgres. It defines two functions, `evaluate_classification` and `evaluate_detection`, which take `GroundTruth` and `Prediction` objects and produce an `Evaluation` object that's equivalent to what's created from the API + client today.
- Add the ability to precompute IOUs and run evaluations using `ValorDetectionManager`
- Vectorize backend operations instead of looping over labels / label keys

## Testing
- `evaluate_detection` and `evaluate_classification` pass all integration and functional tests from Valor; these tests are implemented in `core/test/`. Current test coverage is 91%.
- `core/benchmarks/` contains equivalent benchmark scripts to what's used for `valor`
- Two GH workflows were set up to mirror our test coverage and benchmarking checks

## Next Steps in Future PRs
- Implement `ValorClassificationManager`
- Implement `evaluate_segmentation` and `ValorSegmentationManager`

### Passing Lists of GroundTruth and Prediction Objects

The first way to use `valor_core` is to pass a list of groundtruth and prediction objects to an `evaluate_...` function, like so:

```python

groundtruths = [
    schemas.GroundTruth(
            datum=img1,
            annotations=...
     ), …
]
predictions = [
    schemas.Prediction(
            datum=img1,
            annotations=...
     ), …
]

evaluation = evaluate_detection(
        groundtruths=groundtruths,
        predictions=predictions,
        metrics_to_return=[
            enums.MetricType.AP,
            enums.MetricType.AR,
            enums.MetricType.mAP,
            enums.MetricType.APAveragedOverIOUs,
            enums.MetricType.mAR,
            enums.MetricType.mAPAveragedOverIOUs,
            enums.MetricType.PrecisionRecallCurve,
            enums.MetricType.DetailedPrecisionRecallCurve,
        ],
        pr_curve_iou_threshold=0.5,
        pr_curve_max_examples=1,
    )
```

### Passing DataFrames

The second way to use `valor_core` is to pass in a dataframe of groundtruths and predictions:

```python

groundtruth_df = pd.DataFrame(
        [
            {
                "datum_id": 1,
                "datum_uid": "uid1",
                "id": 1,
                "annotation_id": 1,
                "label_id": 1,
                "label_key": "k1",
                "label_value": "v1",
                "is_instance": True,
                "grouper_key": "k1",
                "polygon": schemas.Polygon.from_dict(
                    {
                        "type": "Polygon",
                        "coordinates": [
                            [[10, 10], [60, 10], [60, 40], [10, 40], [10, 10]]
                        ],
                    }
                ),
                "raster": None,
                "bounding_box": None,
            },
            {
                "datum_id": 1,
                "datum_uid": "uid1",
                "id": 2,
                "annotation_id": 2,
                "label_id": 2,
                "label_key": "k2",
                "label_value": "v2",
                "is_instance": True,
                "grouper_key": "k2",
                "polygon": schemas.Polygon.from_dict(
                    {
                        "type": "Polygon",
                        "coordinates": [
                            [
                                [87, 10],
                                [158, 10],
                                [158, 820],
                                [87, 820],
                                [87, 10],
                            ]
                        ],
                    }
                ),
                "raster": None,
                "bounding_box": None,
            },
            {
                "datum_id": 2,
                "datum_uid": "uid2",
                "id": 3,
                "annotation_id": 3,
                "label_id": 1,
                "label_key": "k1",
                "label_value": "v1",
                "is_instance": True,
                "grouper_key": "k1",
                "polygon": schemas.Polygon.from_dict(
                    {
                        "type": "Polygon",
                        "coordinates": [
                            [[15, 0], [70, 0], [70, 20], [15, 20], [15, 0]]
                        ],
                    }
                ),
                "raster": None,
                "bounding_box": None,
            },
        ]
)
prediction_df = pd.DataFrame(
    [
        {
            "id": 1,
            "annotation_id": 4,
            "score": 0.3,
            "datum_id": 1,
            "datum_uid": "uid1",
            "label_id": 1,
            "label_key": "k1",
            "label_value": "v1",
            "is_instance": True,
            "grouper_key": "k1",
            "polygon": schemas.Polygon.from_dict(
                {
                    "type": "Polygon",
                    "coordinates": [
                        [[10, 10], [60, 10], [60, 40], [10, 40], [10, 10]]
                    ],
                }
            ),
            "raster": None,
            "bounding_box": None,
        },
        {
            "id": 2,
            "annotation_id": 5,
            "score": 0.98,
            "datum_id": 2,
            "datum_uid": "uid2",
            "label_id": 2,
            "label_key": "k2",
            "label_value": "v2",
            "is_instance": True,
            "grouper_key": "k2",
            "polygon": schemas.Polygon.from_dict(
                {
                    "type": "Polygon",
                    "coordinates": [
                        [[15, 0], [70, 0], [70, 20], [15, 20], [15, 0]]
                    ],
                }
            ),
            "raster": None,
            "bounding_box": None,
        },
    ]
)

evaluation = evaluate_detection(
        groundtruths=groundtruth_df,
        predictions=prediction_df,
        metrics_to_return=[
            enums.MetricType.AP,
            enums.MetricType.AR,
            enums.MetricType.mAP,
            enums.MetricType.APAveragedOverIOUs,
            enums.MetricType.mAR,
            enums.MetricType.mAPAveragedOverIOUs,
            enums.MetricType.PrecisionRecallCurve,
            enums.MetricType.DetailedPrecisionRecallCurve,
        ],
        pr_curve_iou_threshold=0.5,
        pr_curve_max_examples=1,
    )
```

## Using a Data Manager

Finally, you can use a manager class (i.e., `ValorDetectionManager`) to run your evaluation. The advantage to using a manager class is a) you won't have to keep all annotation types in memory in a large list and b) we can pre-compute certain columns (i.e., `iou`) in advance of the `.evaluate()` call.

```python
manager = valor_core.ValorDetectionManager(...)
img1 = schemas.Datum(
        uid="uid1",
        metadata={
            "height": image_height,
            "width": image_width,
        },
    )
groundtruths = [
    schemas.GroundTruth(
            datum=img1,
            annotations=...
     ), …
]
predictions = [
    schemas.Prediction(
            datum=img1,
            annotations=...
     ), …
]


# the user passes a list of all groundtruths and predictions for a list of datums
# this allows us to precompute IOUs at the datum_uid + label_key level
manager.add_data(groundtruths=groundtruths, predictions=predictions)

# the user calls .evaluate() to compute the evaluation
evaluation = manager.evaluate()

# the user must pass all groundtruths and predictions for a given datum at once
# this restriction makes it so we can compute IOUs right away and throw away excess info like rasters, saving a significant amount of memory
    manager.add_data_for_datum(groundtruths=groundtruths, predictions=predictions) # throws error since img1 has already been added to the manager's data

# the user must also specify the label map, `convert_annotation_to_type`, etc. when instantiating the object
# once set, these attributes can't be changed since subsequent IOU calculations will become apples-to-oranges with prior calculations
    manager.label_map = some_label_map # throws an error since label map can't be changed, only instantiated
```


api/valor_api/backend/metrics/detection.py
```diff
@@ -645,6 +655,615 @@ def _compute_detailed_curves(
...
+        models.Annotation.datum_id.label("datum_id"),
+        models.Datum.uid.label("datum_uid"),
+        case(
+            grouper_mappings["label_id_to_grouper_id_mapping"],
+            value=models.GroundTruth.label_id,
+        ).label("label_id_grouper"),
+        _annotation_type_to_geojson(annotation_type, models.Annotation).label(
+            "geojson"
+        ),
+        gfunc.ST_AsPNG(models.Annotation.raster).label("raster"),
```
czaloom, line 741: What happens if the annotation is a polygon or bounding box? Its not guaranteed (or that common) to have object detections defined by rasters.

ntlind: then this column would be None, right? there are other tests that check for non-raster cases

core/valor_core/classification.py
```diff
@@ -0,0 +1,1343 @@
+import time
+from collections import defaultdict
+from typing import Dict, List, Optional, Tuple, Union
+
+import numpy as np
+import pandas as pd
+from valor_core import enums, metrics, schemas, utilities
+
+
+def _create_classification_grouper_mappings(
```
czaloom, line 10: Are we switching back to label mapping within the computation?

I was under the impression we were happy with it residing in the data aggregation step since the computation is unnecessarily complicated by it.

czaloom: The data aggregation approach replaces the label  id, key, value in-place so the computation never has to care what the label was originally.

core/valor_core/classification.py
```diff
@@ -0,0 +1,1343 @@
...
+        )
+        .size()
+    )
+
+    cm_counts_df["true_positive_flag"] = (
+        cm_counts_df["pd_grouper_value"] == cm_counts_df["gt_grouper_value"]
+    )
+
+    # resolve pandas typing error
+    assert isinstance(cm_counts_df, pd.DataFrame)
```
czaloom, line 81: `assert` should be replaced with a if-statement that raises. 

See [reference](https://www.reddit.com/r/learnpython/comments/zd217y/is_using_assert_considered_good_practice/).

czaloom: this is probably more of a nit since this condition is for typing errors and doesn't affect runtime.

core/valor_core/classification.py
```diff
@@ -0,0 +1,1343 @@
...
+            as_index=False,
+        )
+        .max()
+    )
+
+    # catch pandas typing error
+    if not isinstance(prediction_df, pd.DataFrame) or not isinstance(
+        max_scores_by_grouper_key_and_datum_id, pd.DataFrame
+    ):
+        raise ValueError
```
czaloom, line 404: nit: Could you add an error message?

core/valor_core/classification.py
```diff
@@ -0,0 +1,1343 @@
...
+    )
+
+    metrics_to_output += _calculate_accuracy_metrics(cm_counts_df=cm_counts_df)
+
+    metrics_to_output += _calculate_rocauc(
+        prediction_df=prediction_df, groundtruth_df=groundtruth_df
+    )
+
+    # handle type error
+    assert metrics_to_return
```
czaloom, line 1218: nit: use `raise` instead of `assert`. See [reference](https://www.reddit.com/r/learnpython/comments/zd217y/is_using_assert_considered_good_practice/).

core/valor_core/classification.py
```diff
@@ -0,0 +1,1343 @@
...
+    ]
+    confusion_matrices = [cm.to_dict() for cm in confusion_matrices]
+
+    return confusion_matrices, metrics_to_output
+
+
+def evaluate_classification(
+    groundtruths: Union[pd.DataFrame, List[schemas.GroundTruth]],
+    predictions: Union[pd.DataFrame, List[schemas.Prediction]],
+    label_map: Optional[Dict[schemas.Label, schemas.Label]] = None,
```
czaloom, line 1241: Expanding on my comment above, i feel that label mapping should occur in the data aggregation layer and not within the computation itself. This will reduce a lot of complexity in the computation and allow for tighter testing coverage.

core/valor_core/detection.py
```diff
@@ -0,0 +1,1470 @@
...
+
+        joint_df = joint_df.join(iou_calculation_df["iou_"])
+
+    else:
+        iou_calculation_df = joint_df.loc[
+            ~joint_df["converted_geometry_gt"].isnull()
+            & ~joint_df["converted_geometry_pd"].isnull(),
+            ["converted_geometry_gt", "converted_geometry_pd"],
+        ].apply(
+            lambda row: geometry.calculate_bbox_iou(
```
czaloom, line 132: @ekorman since we are losing the ability to compute IoU for polygons could we reconsider a library like [shapely](https://shapely.readthedocs.io/en/stable/manual.html)?

core/valor_core/detection.py
```diff
@@ -0,0 +1,1470 @@
...
+        .groupby("label_id_grouper", as_index=False)["recall_for_AR"]
+        .mean()
+    )
+
+    # add back "label" after grouping operations are complete
+    ar_metrics_df["label"] = ar_metrics_df["label_id_grouper"].map(
+        grouper_mappings["grouper_id_to_grouper_label_mapping"]
+    )
+
+    # resolve typing error
```
czaloom, line 431: nit: stray comment

core/valor_core/detection.py
```diff
@@ -0,0 +1,1470 @@
...
+        for key, value in curves.items()
+    ]
+
+
+def _add_samples_to_dataframe(
+    detailed_pr_curve_counts_df: pd.DataFrame,
+    detailed_pr_calc_df: pd.DataFrame,
+    max_examples: int,
+    flag_column: str,
+):  # -> pd.DataFrame:
```
czaloom, line 594: nit: commented code

core/valor_core/detection.py
```diff
@@ -0,0 +1,1470 @@
...
+            value=dict(value),
+            pr_curve_iou_threshold=pr_curve_iou_threshold,
+        )
+        for key, value in detailed_pr_curves.items()
+    ]
+
+    return detailed_pr_metrics
+
+
+def _create_detection_grouper_mappings(
```
czaloom, line 1103: see comments above on label mapping

core/valor_core/geometry.py
```diff
@@ -0,0 +1,1440 @@
...
+import json
+from base64 import b64decode, b64encode
+from dataclasses import dataclass
+from typing import Any, Optional, TypedDict, Union
+
+import numpy as np
+import PIL.Image
+
+
+def calculate_bbox_intersection(bbox1: np.ndarray, bbox2: np.ndarray) -> float:
```
czaloom, line 11: note - This works for axis-aligned bounding boxes but we lose support for anything transformed.

core/valor_core/geometry.py
czaloom, line 1: nit - This file seems to be a mix of classes and functions from various parts of the API. 

Should we organize valor-core in a similar fashion to separate concepts better? For example, Im not sure if embeddings belong in `geometry.py`. Also, the definition of the box IoU functions that take `np.ndarray` as input is a bit weird given that the `Box` schema is defined within the same file and is incompatible.

core/valor_core/metrics.py
```diff
@@ -0,0 +1,666 @@
...
+    """
+
+    __type__ = "PrecisionRecallCurve"
+    value: Dict[
+        str,
+        Dict[
+            float,
+            Dict[str, Optional[Union[int, float]]],
+        ],
+    ]
```
czaloom, line 481: nit: Could you add comments defining what these nested structures are?

core/valor_core/schemas.py
czaloom, line 1: There are schemas in `geometry.py` that could probably be moved here to group related structures.

core/valor_core/classification.py
ntlind: yes, the core reason being that pandas requires me to drop the mapping columns during some `group_by` operations, so I need a way to map them back after the `group_by` is done.

core/valor_core/utilities.py
```diff
@@ -0,0 +1,848 @@
+from typing import Dict, List, Optional, Union
+
+import numpy as np
+import pandas as pd
+from valor_core import enums, geometry, schemas
+
+
+def validate_label_map(
```
czaloom, line 8: See comments above on label mapping.

core/valor_core/detection.py
ntlind: this will be added in a future PR; let's discuss it then (see 'Issues to Fix' in PR notes)

core/valor_core/classification.py
ntlind: see above

core/valor_core/schemas.py
ntlind: can you go into more detail? I tried to move the minimum number of objects over; what's missing?

core/valor_core/utilities.py
```diff
@@ -0,0 +1,848 @@
...
+    if not np.any(rows) or not np.any(cols):
+        raise ValueError("Raster is empty, cannot create bounding box.")
+
+    ymin, ymax = np.where(rows)[0][[0, -1]]
+    xmin, xmax = np.where(cols)[0][[0, -1]]
+
+    return geometry.Box.from_extrema(xmin, xmax + 1, ymin, ymax + 1)
+
+
+def _convert_raster_to_polygon(raster: np.ndarray) -> geometry.Polygon:
```
czaloom, line 616: Is there a reference for the algorithm used here?

core/valor_core/utilities.py
```diff
@@ -0,0 +1,848 @@
...
+                if isinstance(x, geometry.Polygon)
+                else None
+            )  # pyright: ignore
+        )
+
+    elif target_type == enums.AnnotationType.BOX:
+        df["converted_geometry"] = df["converted_geometry"].map(
+            lambda x: (
+                _convert_raster_to_box(
+                    x.to_array()  # pyright: ignore
```
czaloom, line 761: nit - could you make these ignores more descriptive?

core/valor_core/schemas.py
czaloom: Sorry, I forgot to referenced the comment above talking about how `geometry.py` seems to be a mixture of functionality. There are schemas and functions that do not interact stored in the same place.

core/valor_core/utilities.py
ntlind: there isn't. I need to test this further when adding better conversion tests see ("Issues to fix..")

core/valor_core/classification.py
ntlind: fixed

core/valor_core/geometry.py
ntlind: generally, I'm down to re-organize where you have specific suggestions. I agree that Embedding should live in schemas; thanks for that suggestion!

> Also, the definition of the box IoU functions that take np.ndarray as input is a bit weird given that the Box schema is defined within the same file and is incompatible.

those functions (like `calculate_bbox_intersection`) seem "geometry" related to me. we could move them into a separate file such that functions and classes are separate, but I don't see a big benefit there

core/valor_core/schemas.py
ntlind: thanks for clarifying; responded above

core/valor_core/classification.py
ntlind: I can work on this. the data aggregation change happened after I'd already finished this part of the code

core/valor_core/geometry.py
czaloom: I agree that the calculation makes sense living there, its mostly that all the geometric schemas don't interact with the geometric functions due to their differing purposes (ingest vs computation). To me this means they should be stored separately.

czaloom: The simplest solution is to just have all schemas live in `schemas.py`

ekorman:
can you add a README.md in `core/` with a brief description of this package?

core/valor_core/classification.py
ntlind: made this change. thanks for the suggestion!

core/valor_core/utilities.py
```diff
@@ -0,0 +1,945 @@
...
+
+
+def _add_converted_geometry_column(
+    df: pd.DataFrame,
+    target_type: enums.AnnotationType,
+) -> pd.DataFrame:
+    """
+    Add a column with converted geometries to the DataFrame.
+
+    The function checks that each annotation contains only one type of geometry
```
czaloom, line 799: Why are we regressing on functionality? An annotation can include multiple representations of itself in the current backend.

YOLO inferences for example provides both bitmasks and bounding boxes per annotation.

core/tests/conftest.py
```diff
@@ -0,0 +1,2940 @@
...
+        )
+        for i in range(6)
+    ]
+
+    return (gts, preds)
+
+
+@pytest.fixture
+def classification_functional_test_groundtruth_df():
+    """Used in test_rocauc_with_label_map so that we can test _calculate_rocauc directly, since this original text violated the matching groundtruth/prediction label keys criteria."""
```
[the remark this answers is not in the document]
ntlind: the goal of this PR is to make sure that `valor_core` passes all functional and integration tests from `valor`, so yes

core/valor_core/utilities.py
ntlind: there are no integration or functional tests where this is true, so this is part of the "blind risk" mentioned in "Issues to Fix..". I can consider relaxing this assumption when implementing better tests in the next PR

core/valor_core/classification.py
```diff
@@ -0,0 +1,1291 @@
...
+                (chain_df["label_value_pd"] == chain_df["label_value_gt"])
+            )
+        )
+        .drop(
+            columns=[
+                "annotation_id_gt",
+                "created_at_gt",
+                "annotation_id_pd",
+                "created_at_pd",
+                "label_pd",
```
rsbowman-striveworks, line 706: Probably can expand this list of cols to ignore, or just specify the columns you want to _keep_ might be even better

core/valor_core/classification.py
```diff
@@ -0,0 +1,1291 @@
...
+    if not groundtruths_associated_with_true_positives.empty:
+        groundtruths_associated_with_true_positives.columns = [
+            "confidence_threshold",
+            "groundtruths_associated_with_true_positives",
+        ]
+        pr_calc_df = pr_calc_df.merge(
+            groundtruths_associated_with_true_positives,
+            on=["confidence_threshold"],
+            how="left",
+        )
```
rsbowman-striveworks, line 755: Here and below this list gets added to every row of a pretty big dataframe, and I believe the row only depends on `confidence_threshold`.   Might make sense to break up the computation, loop over threshold values in some higher function and put together what you need afterward.  That lets you avoid putting these lists in the dataframe.

Also consider doing each `label_key` separately, I think each of these is completely independent of the others.

Separating the computation by label key and threshold lets you avoid the outer join with label key and the similar blowup with confidence threshold on line 715.

core/valor_core/classification.py
```diff
@@ -0,0 +1,1291 @@
...
+
+    return pr_curve_counts_df
+
+
+def _calculate_pr_curves(
+    prediction_df: pd.DataFrame,
+    groundtruth_df: pd.DataFrame,
+    metrics_to_return: list,
+    pr_curve_max_examples: int,
+) -> list[metrics.PrecisionRecallCurve]:
```
rsbowman-striveworks, line 679: Might make sense in this function to use some kind of "label ID" instead of a string `label_value`.  The strings are redundant and can use lots of memory if they're big.  Map to an integer and then back again at the very end?

core/valor_core/classification.py
ntlind: all the other columns are used in some way; I've already tried to remove others

core/valor_core/classification.py
ntlind: label tuples are significantly easier to debug than label IDs. I don't think the trade-off is worth it here as these strings are usually very small

core/valor_core/classification.py
ntlind: we're purposefully only doing array operations here (not mapping with lists or separating by label key) to improve runtime performance and debug-ability. your suggestions make sense, but require trading peak memory usage for speed.  I'm considering this a nit until we get real requirements that would justify this work.

core/tests/functional-tests/test_detection.py
```diff
@@ -0,0 +1,4434 @@
...
+        metrics_to_return=selected_metrics,
+    )
+
+    metrics = eval_job.metrics
+    assert set([metric["type"] for metric in eval_job.metrics]) == set(
+        selected_metrics
+    )
+
+
+def test_two_groundtruths_one_datum(
```
ntlind, line 4264: note that this test shows that you can have multiple GroundTruths per datum without issue. in the API, this operation will throw an error since postgres will try to create the same datum twice (which makes sense). this issue doesn't exist in `valor_core`, so having two GroundTruths with one annotation each is equivalent to having one GroundTruth with two annotations. 

calling this out in case you all want to discuss further. I don't think there's a need to implement another validation check to disallow this behavior, but open to your thoughts

core/valor_core/classification.py
rsbowman-striveworks: When I look at the dataframe here I see columns like `bounding_box_gt` and `polygon_gt`, surely those aren't needed in this function?

ntlind: you're right: I thought you were talking about the columns that get dropped for detection tasks. just made this change; thanks for the suggestion

core/valor_core/classification.py
```diff
@@ -0,0 +1,1321 @@
...
+            .to_dict()
+        )
+
+        pr_calc_df["false_negative_flag"] = pr_calc_df.apply(
+            lambda row: row["id_gt"]
+            not in confidence_interval_to_true_positive_groundtruth_ids_dict.get(
+                row["confidence_threshold"], set()
+            ),
+            axis=1,
+        )
```
rsbowman-striveworks, line 772:
```suggestion
        false_negative_mask = pd.Series(False, index=pr_calc_df.index)
        for threshold, elements in confidence_interval_to_true_positive_groundtruth_ids_dict.items():
            threshold_mask = pr_calc_df["confidence_threshold"] == threshold
            membership_mask = ~pr_calc_df["id_gt"].isin(elements)
            false_negative_mask |= threshold_mask & membership_mask
        pr_calc_df["false_negative_flag"] = false_negative_mask
```

After the 9e606cd changes, this line and the corresponding one below account for a large majority of the running time of this function.   I found a large speedup (and decreased memory usage) by using array functions as suggested.

core/valor_core/classification.py
```diff
@@ -0,0 +1,1321 @@
...
+        pr_calc_df["no_predictions_false_negative_flag"] = (
+            pr_calc_df.apply(
+                lambda row: row["id_gt"]
+                not in confidence_interval_to_misclassification_fn_groundtruth_ids_dict.get(
+                    row["confidence_threshold"], set()
+                ),
+                axis=1,
+            )
+            & pr_calc_df["false_negative_flag"]
+        )
```
rsbowman-striveworks, line 812:
```suggestion
        no_prediction_false_negative_mask = pd.Series(False, index=pr_calc_df.index)
        for threshold, elements in confidence_interval_to_misclassification_fn_groundtruth_ids_dict.items():
            threshold_mask = pr_calc_df["confidence_threshold"] == threshold
            membership_mask = pr_calc_df["id_gt"].isin(elements)
            no_prediction_false_negative_mask |= threshold_mask & membership_mask
        pr_calc_df["no_predictions_false_negative_flag"] = (
            ~no_prediction_false_negative_mask & pr_calc_df["false_negative_flag"]
        )
```
This is the corresponding change to the line 766 one.

core/valor_core/geometry.py
```diff
@@ -0,0 +1,137 @@
...
+    bbox : List[Tuple[float, float]]
+        Coordinates of the bounding box.
+
+    Returns
+    ----------
+    bool
+        True if the bounding box is axis-aligned, otherwise False.
+    """
+    return all(
+        x1 == x2 or y1 == y2
```
czaloom, line 81: This should be an `and` to enforce alignment on all edges.

core/valor_core/geometry.py
```diff
@@ -0,0 +1,137 @@
...
+    vectors = []
+    for (x1, y1), (x2, y2) in zip(bbox, bbox[1:] + bbox[:1]):
+        vectors.append(np.array([x2 - x1, y2 - y1]))
+
+    angles = [
+        angle_between(vectors[i], vectors[(i + 1) % len(vectors)])
+        for i in range(len(vectors))
+    ]
+
+    return not all(
```
czaloom, line 116: Why not just use `not is_axis_aligned`?

core/valor_core/geometry.py
```diff
@@ -0,0 +1,137 @@
...
+    bool
+        True if the bounding box is axis-aligned, otherwise False.
+    """
+    return all(
+        x1 == x2 or y1 == y2
+        for (x1, y1), (x2, y2) in zip(bbox, bbox[1:] + bbox[:1])
+    )
+
+
+def is_skewed(bbox: List[Tuple[float, float]]) -> bool:
```
czaloom, line 86: Why not just use `not is_axis_aligned`?

core/valor_core/managers.py
```diff
@@ -0,0 +1,352 @@
...
+            enums.MetricType.mAPAveragedOverIOUs,
+        ]
+    )
+    iou_thresholds_to_compute: list[float] = field(
+        default_factory=lambda: [round(0.5 + 0.05 * i, 2) for i in range(10)]
+    )
+    iou_thresholds_to_return: list[float] = field(
+        default_factory=lambda: [0.5, 0.75]
+    )
+    recall_score_threshold: float = field(default_factory=lambda: 0.0)
```
czaloom, line 84: nit: field has a arg called default 

ex. `field(default=0.0)`

czaloom: I would prefer to use this over a lambda when possible.

api/valor_api/backend/metrics/classification.py
czaloom, line 1: Why make these API changes?

core/README.md
```diff
@@ -0,0 +1,243 @@
...
+            {
+                "datum_id": 1,
+                "datum_uid": "uid1",
+                "id": 1,
+                "annotation_id": 1,
+                "label_id": 1,
+                "label_key": "k1",
+                "label_value": "v1",
+                "is_instance": True,
+                "grouper_key": "k1",
```
czaloom, line 70: What is `grouper_key` in this context?

core/README.md
```diff
@@ -0,0 +1,243 @@
...
+            {
+                "datum_id": 1,
+                "datum_uid": "uid1",
+                "id": 2,
+                "annotation_id": 2,
+                "label_id": 2,
+                "label_key": "k2",
+                "label_value": "v2",
+                "is_instance": True,
+                "grouper_key": "k2",
```
czaloom, line 91: See comment above on `grouper_key`

core/README.md
```diff
@@ -0,0 +1,243 @@
...
+            {
+                "datum_id": 2,
+                "datum_uid": "uid2",
+                "id": 3,
+                "annotation_id": 3,
+                "label_id": 1,
+                "label_key": "k1",
+                "label_value": "v1",
+                "is_instance": True,
+                "grouper_key": "k1",
```
czaloom, line 118: See above comment on `grouper_key`

core/README.md
czaloom: It is used in all examples below too

core/tests/functional-tests/test_classification.py
```diff
@@ -0,0 +1,1385 @@
...
+    pred_clfs_tabular: list[list[float]],
+):
+    assert len(gt_clfs_tabular) == len(pred_clfs_tabular)
+
+    groundtruths = []
+    predictions = []
+    for i, label_value in enumerate(gt_clfs_tabular):
+        if (
+            i % 3 == 2
+        ):  # core doesn't use filters, so this emulates the results of the original test
```
czaloom, line 347: nit: We can just remove this test since its filter related.

core/tests/functional-tests/test_classification.py
```diff
@@ -0,0 +1,1385 @@
...
+                }
+            ],
+        }
+        # other label keys not included for testing purposes
+    ]
+
+    eval_job = evaluate_classification(
+        groundtruths=gt_clfs_with_label_maps,
+        predictions=pred_clfs_with_label_maps,
+        label_map=label_mapping,
```
czaloom, line 656: I thought we agreed that label mapping was a part of the aggregation layer?

core/tests/functional-tests/test_classification.py
```diff
@@ -0,0 +1,1385 @@
...
+    )
+
+    computed_metrics = evaluation.metrics
+
+    assert all([metric["value"] == 0 for metric in computed_metrics])
+    assert all([metric in computed_metrics for metric in expected_metrics])
+    assert all([metric in expected_metrics for metric in computed_metrics])
+
+
+def test_compute_confusion_matrix_at_grouper_key_using_label_map(
```
czaloom, line 919: See above on label mapping.

core/valor_core/managers.py
ntlind: thanks for the suggestion! changed this for all non-collections

core/README.md
ntlind: removed all grouper_ references

core/tests/functional-tests/test_classification.py
ntlind: deleted

api/valor_api/backend/metrics/classification.py
ntlind: these were just leftover linter changes. removed them

core/valor_core/classification.py
czaloom: @rsbowman-striveworks I agree with your suggestion but this is something we would likely implement in a follow-up PR. Would you be able to add this to your memory ticket?

core/valor_core/managers.py
```diff
@@ -0,0 +1,322 @@
...
+@dataclass
+class ValorDetectionManager:
+    """
+    Manages the evaluation of object detection predictions against groundtruths.
+
+    Attributes
+    ----------
+    datum_uids : set
+        A set of unique identifiers for the data samples.
+    label_map : dict
```
czaloom, line 37: Could you update this to be `dict[schemas.Label, schemas.Label]`

czaloom: And datum_uids to `set[str]`

czaloom: And convert_annotations_to_type to `AnnotationType, optional` to match Numpy style

core/valor_core/metrics.py
```diff
@@ -0,0 +1,669 @@
...
+    Attributes
+    ----------
+    label_key : str
+        A label for the matrix.
+    entries : List[ConfusionMatrixEntry]
+        A list of entries for the matrix.
+
+    Attributes
+    ----------
+    matrix : np.zeroes
```
czaloom, line 646: Why is the type `np.zeroes`?

core/valor_core/schemas.py
```diff
@@ -0,0 +1,1739 @@
...
+
+
+@dataclass
+class Annotation:
+    """
+    A class used to annotate `GroundTruths` and `Predictions`.
+
+    Attributes
+    ----------
+    metadata: Dictionary
```
czaloom, line 1351: Could you change this to `dict[str, Any]`

core/valor_core/schemas.py
```diff
@@ -0,0 +1,1739 @@
...
+@dataclass
+class Datum:
+    """
+    A class used to store information about a datum for either a 'GroundTruth' or a 'Prediction'.
+
+    Attributes
+    ----------
+    uid : String
+        The UID of the datum.
+    metadata : Dictionary
```
czaloom, line 1221: I think you are using the symbolic types for the docstrings in this file. Could you update with proper typing?

core/valor_core/utilities.py
```diff
@@ -0,0 +1,1013 @@
...
+    DataFrames based on the provided label map. If the `label_map` is not provided, the function returns the original DataFrames
+    without modification.
+
+    Parameters
+    ----------
+    groundtruth_df : pd.DataFrame
+        DataFrame containing groundtruth data with columns `label_key`, `label_value`, and `label_id`.
+    prediction_df : pd.DataFrame
+        DataFrame containing prediction data with columns `label_key`, `label_value`, and `label_id`.
+    label_map : Optional[schemas.Dict[schemas.Label, schemas.Label]]
```
czaloom, line 26: Could you update this to `dict[schemas.Label, schemas.Label], optional`?

core/valor_core/utilities.py
```diff
@@ -0,0 +1,1013 @@
...
+    """
+    Validates that every datum has the same set of label keys for both ground truths and predictions. This check is only needed for classification tasks.
+
+    Parameters
+    ----------
+    groundtruths : pd.DataFrame
+        The DataFrame containing ground truth data.
+    predictions : pd.DataFrame
+        The DataFrame containing prediction data.
+    label_map : Optional[Dict[schemas.Label, schemas.Label]]
```
czaloom, line 238: nit: Could you update this to: `Dict[schemas.Label, schemas.Label], optional`

core/tests/functional-tests/test_detection_manager.py
czaloom, line 1: Are these the same tests as in `test_detection.py`? Or are these drawn from the integration tests?

ntlind: they are mostly the same tests used in test_detection.py, but they use the DetectionManager instead of `evaluate_detection`

czaloom (approved):

PR #654, opened 2024-07-03
Integration Test Failing due to Unordered List

czaloom:
### Reproducible Example

```python
`integration_tests/client/datasets/test_dataset.py::test_create_tabular_dataset_and_add_groundtruth`
```

### Issue Description

The test queries from postgres and treats the response as an ordered list which it is not.

Test failure is dependent on what order postgres returns its response.

### Expected Behavior

The test should not assume an ordered response from postgres.

### DIFF
--- a/integration_tests/client/datasets/test_dataset.py
+++ b/integration_tests/client/datasets/test_dataset.py
@@ -272,17 +272,13 @@ def test_create_tabular_dataset_and_add_groundtruth(
     assert set(d.uid for d in data) == {"uid1", "uid2"}
 
     # check metadata is there
-    metadata_links = data[0].meta
-    assert len(metadata_links) == 1
-    assert "metadatum1" in metadata_links
-    assert metadata_links["metadatum1"] == "temporary"
-
-    metadata_links = data[1].meta
-    assert len(metadata_links) == 2
-    assert "metadatum2" in metadata_links
-    assert metadata_links["metadatum2"] == "a string"
-    assert "metadatum3" in metadata_links
-    assert metadata_links["metadatum3"] == 0.45
+    for datum in data:
+        if "metadatum1" in datum.meta:
+            assert len(datum.meta) == 1
+            assert datum.meta == md1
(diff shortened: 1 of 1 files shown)


integration_tests/client/datasets/test_dataset.py
```diff
@@ -272,17 +272,15 @@ def test_create_tabular_dataset_and_add_groundtruth(
     assert set(d.uid for d in data) == {"uid1", "uid2"}
 
     # check metadata is there
-    metadata_links = data[0].meta
-    assert len(metadata_links) == 1
-    assert "metadatum1" in metadata_links
-    assert metadata_links["metadatum1"] == "temporary"
```
ntlind, line 278: can you add back this assertion to make sure this change equivalent to the old test?

integration_tests/client/datasets/test_dataset.py
```diff
@@ -272,17 +272,15 @@ def test_create_tabular_dataset_and_add_groundtruth(
...
-
-    metadata_links = data[1].meta
-    assert len(metadata_links) == 2
-    assert "metadatum2" in metadata_links
-    assert metadata_links["metadatum2"] == "a string"
```
ntlind, line 283: same with this line and 285

integration_tests/client/datasets/test_dataset.py
czaloom: The value isnt defined in the test so im not sure why this is necessary. The assertion is handled by `datum.meta == md1` or `datum.meta == md2`

integration_tests/client/datasets/test_dataset.py
czaloom: see above.

integration_tests/client/datasets/test_dataset.py
czaloom: There is no regression in test coverage so im confused why this matters.

ntlind: makes sense!

ntlind (approved):

PR #660, opened 2024-07-08
Expanded Client Timeout Controls

czaloom:
## Changes
- added timeout controls to all Connection memebers
- added default timeout of 30s for GET, POST, PUT, DELETE

### DIFF
--- a/client/valor/client.py
+++ b/client/valor/client.py
@@ -89,6 +89,32 @@ def get_json_size(json_obj: object, encoding: str = "utf-8") -> int:
     )
 
 
+def _format_request_timeout(
+    timeout: Optional[float], default: float
+) -> Optional[float]:
+    """
+    Converts user-requested timeout into requests library format.
+
+    Parameters
+    ----------
+    timeout : float, optional
+        The user requested timeout in seconds. Timeout <= 0 disables the timeout.
+    default : float
+        The default timeout in seconds. Used when user timeout is None.
+
+    Returns
+    -------
+    float | None
+        The validated request timeout.
+    """
+    if timeout is None:
+        return default
+    elif timeout <= 0:
+        return None
+    else:
+        return timeout
+
+
 @dataclass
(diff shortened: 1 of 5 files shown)


client/valor/client.py
```diff
@@ -89,6 +89,32 @@ def get_json_size(json_obj: object, encoding: str = "utf-8") -> int:
     )
 
 
+def _format_request_timeout(
+    timeout: Optional[float], default: float
+) -> Optional[float]:
+    """
+    Converts user-requested timeout into requests library format.
```
ntlind, line 96:
```suggestion
    Converts user-requested timeout into requests library format to avoid issues with passing `None` to `requests.timeout`.
```

client/valor/coretypes.py
```diff
@@ -1761,7 +1815,7 @@ def delete_evaluation(self, evaluation_id: int, timeout: int = 0) -> None:
         ----------
         evaluation_id : int
             The id of the evaluation to be deleted.
-        timeout : int
+        timeout : int, default = 1
```
ntlind, line 1818: is this default right?

client/valor/coretypes.py
```diff
@@ -1761,7 +1815,7 @@ def delete_evaluation(self, evaluation_id: int, timeout: int = 0) -> None:
...
             The number of seconds to wait in order to confirm that the model was deleted.
         """
         self.conn.delete_evaluation(evaluation_id)
```
ntlind, line 1821: is timeout used here?

client/valor/coretypes.py
czaloom: Good catch, changed it to `0`.

client/valor/coretypes.py
czaloom: Its carried over from a previous pr. This timeout is used for polling.

ntlind (approved):

PR #673, opened 2024-07-17
Evaluation Data Aggregation

czaloom:
# Changes
- separated data aggregation from evaluation computation
- unified label mapping across all evaluation types
- ROCAUC fixed in classification integration test with label maps

## Label Mapping

Valor now does the following when a label map is recieved.
1. New labels are added to the database.
2. Mapping is converted into a label_id --> label_id dictionary.

This dictionary is used in a case statement in the data aggregator to override the existing labels.


api/tests/functional-tests/backend/metrics/test_segmentation.py
```diff
@@ -53,146 +52,6 @@ def _create_data(
     )
 
 
-def test_query_generators(
```
czaloom, line 56: removed b/c this function no longer exists

api/valor_api/backend/metrics/classification.py
```diff
@@ -313,31 +559,31 @@ def _compute_curves(
...
+        for label_key, value in pr_output.items()
+    ]
 
-    if enums.MetricType.DetailedPrecisionRecallCurve in metrics_to_return:
-        output += [
-            schemas.DetailedPrecisionRecallCurve(
-                label_key=grouper_key, value=dict(detailed_pr_output)
-            )
-        ]
+    detailed_pr_curves = [
```
ntlind, line 570: are we making it so we always run DetailedPRCurves by removing the wold `if` condition on 324? not sure I'm reading this change correctly.

czaloom: This is just a result aggregator, since `detailed_pr_output` is empty this list comprehension will be skipped.

czaloom: The computation block is earlier in this function and should be controlled by an `if` statement

czaloom: See line 451!

api/valor_api/backend/metrics/classification.py
```diff
@@ -981,26 +1154,173 @@ def _compute_confusion_matrix_and_metrics_at_grouper_key(
...
-    return confusion_matrix, output
+    return [
+        confusion_matrix
+        for confusion_matrix in confusion_matrices.values()
+        if confusion_matrix
+    ], metrics
 
 
-def _compute_clf_metrics(
+def _aggregate_data(
```
ntlind, line 1164: this name isn't descriptive; could we change it back to something to do with computing classification metrics (per the docstring)?

api/valor_api/backend/metrics/classification.py
```diff
@@ -981,26 +1154,173 @@ def _compute_confusion_matrix_and_metrics_at_grouper_key(
...
+            predictions_subquery.c.datum_id,
+            predictions_subquery.c.datum_uid,
+            models.Label.id,
+            models.Label.key,
+            models.Label.value,
+        )
+        .cte()
+    )
+
+    def groundtruth_label_query():
```
ntlind, line 1271: move these functions out or to the top, make private, add docstring?

integration_tests/client/metrics/test_classification.py
```diff
@@ -1101,6 +1101,15 @@ def test_evaluate_classification_with_label_maps(
         threshold,
         metric,
     ), expected_value in pr_expected_values.items():
+        print(
```
ntlind, line 1104: remove print statement?

api/valor_api/backend/metrics/classification.py
czaloom: The function's role is to collect data, It has nothing to do with computing metrics so im a little confused.

Ill update the docstring header since that is wrong!

api/valor_api/backend/metrics/classification.py
czaloom: This is remanent of my profiling, they are now variables!

ntlind (approved):

PR #690, opened 2024-07-30
Improve Text Generation Instructions and Tests

bnativi:
### Improvements to Instructions
The llm instructions for the text generation metrics were copied from DeepEval, with some slight modifications. However the instructions from DeepEval have many issues, including misspellings, grammatical errors, inconsistent formatting, poorly organized examples and inconsistent use of terminology. This PR fixes many of those errors and standardizes the formatting. 

Additionally, the DeepEval instructions request an optional 'reason' field from the LLM for verdicts. This is requested following the 'verdict' depending on the 'verdict'. For example, if the verdict is that there is bias, the llm should give a reason as to why there is bias. Because Valor is not exposing these reasons to the user, they are not currently adding anything to the computations. They also result in extra prompt and completion tokens, increasing the cost for the user.  This PR removes the instructions with the 'reason' fields, and adds two versions of both verdict instruction. One version just removes the reasons from the JSON. The reasons for the examples are just moved to the end of the example, and 'reason' is not requested from the LLM. The quality of the evaluations should be about the same as before, but with slightly lower token cost. The other version changes the 'reason' field to the 'analysis' field, and puts the 'analysis' before the 'verdict'. This is effectively a form of chain of thought prompting. Because the LLM writes an analysis before giving the verdict, the LLM can use its own analysis to better inform its verdict. This increases the token cost, but should improve the quality of the metrics.

On a small initial test of the cost, the version with 'analysis' increased the cost by approximately 27% over the version without 'analysis'. I tested the current 7 LLM guided metrics on 10 RAG datums using GPT-4o and measured the total number of prompt and completion tokens. The respective versions used:
- no analysis
    - 28,522 prompt tokens
    -   1,744 completion tokens
- with analysis
    - 32,315 prompt tokens
    -   3,539 completion tokens
As completion tokens are 3x as expensive as prompt tokens for GPT-4o, the increase in cost is 27%. 

The code is currently set up to use the analysis version. There is not currently any way for the user to specify what version, it is currently hard coded to the analysis version. We could leave both versions in and let the user choose, or we could test the quality of both versions. 

### External Integration Tests

Bias, context relevance, faithfulness, hallucination and toxicity were added to the external integration tests. Each metric has it's own basic test data intended to have "obvious" scores. The test data has been modified and tweaked so that the llms pretty often return the desired score (at least 90% of the time). 

### Terminology

Change "context" to "context_list" where appropriate. "context" is now used only to refer to a single piece of context, not the entire list of contexts. This is consistent with the terminology in RAGAS.

### Code Coverage

Add additional tests to improve the code coverage for text generation.


api/tests/functional-tests/backend/core/test_llm_clients.py
```diff
@@ -122,19 +126,19 @@
...
 
         {
-            "verdict": "no",
-            "reason": "The text and context mention disagree on when Abraham Lincoln was born."
+            "verdict": "yes",
+            "reason": "The text and context disagree on when Abraham Lincoln was born."
         },
         {
-            "verdict": "no",
+            "verdict": "yes",
```
czaloom, line 141: Why did these tests change?

api/valor_api/exceptions.py
```diff
@@ -397,6 +397,14 @@ class InvalidLLMResponseError(Exception):
     pass
 
 
+class BadValueInTestLLMClientsError(Exception):
```
czaloom, line 400: Is this error only used for testing? If so, Im not sure it should be included in the production code.

api/tests/functional-tests/backend/core/test_llm_clients.py
bnativi: I changed these from agreement verdicts to contradiction verdicts, so all the tests needed to swap 'yes' to 'no and vice versa. 

I made this change because DeepEval used an instruction asking for agreement verdicts, but their instruction was very confusing. They tell the LLM to respond 'yes' if the prediction text agrees with the contexts, but 'no' if only if the prediction text directly contradicts the contexts. I don't know if agreement is really the exact opposite of contradiction. I felt like this left the case ambiguous when the text is unrelated to the contexts. 

The hallucination metric is the proportion of contradictions, so I felt like it would be better to just ask the LLM for contradiction verdicts. That's why I flipped from agreement verdicts to contradiction verdicts. 

Unfortunately, this is an artefact of DeepEval having low quality instructions. I would look at RAGAS here if I could, but RAGAS doesn't have a hallucination metric.

api/valor_api/backend/core/llm_clients.py
```diff
@@ -11,6 +11,18 @@
 )
 from pydantic import BaseModel
 
+from valor_api.backend.core.llm_instructions_analysis import (
+    _generate_answer_relevance_verdicts_instruction,
```
ntlind, line 15: should these functions be private if they are imported here?

api/valor_api/backend/core/llm_instructions_analysis.py
```diff
@@ -0,0 +1,573 @@
...
+        The instruction for the llm.
+    """
+    return f"""Based on the text, generate a list of OPINIONS presented in the text. Claims and undisputed truths are NOT opinions.
+
+IMPORTANT: Return in JSON format with the "opinions" key mapping to a list of strings. No words or explanation is needed.
+Cited opinions should NOT be included as they are not opinions of the author of the text.
+Incorrect facts do NOT count as opinions.
+
+===== EXAMPLE ======
+Example Text: "Hitler hated jews, but I think the hate is unwarranted. Fox News thinks Donald Trump is a better President than Joe Biden. Earth is the smallest planet in our solar system."
```
ntlind, line 65: nit: some of these examples are a bit edgier than what I would include in a public repo.

api/valor_api/backend/core/llm_instructions_analysis.py
```diff
@@ -0,0 +1,573 @@
+def _generate_claims_instruction(text: str) -> str:
+    """
+    Instruction template was adapted from DeepEval's codebase https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/faithfulness/template.py.
```
ntlind, line 3: in these docstrings, can you start by describing what this function does / why it's needed?

api/valor_api/backend/core/llm_instructions_analysis.py
```diff
@@ -0,0 +1,573 @@
+def _generate_claims_instruction(text: str) -> str:
```
ntlind, line 1: all of the code in this file is not being used since you decided to go with the non-analysis version, right?

nit: it might make sense to save this code locally and delete it from the repo

api/valor_api/backend/core/llm_instructions_no_analysis.py
```diff
@@ -0,0 +1,560 @@
+def _generate_claims_instruction(text: str) -> str:
+    """
+    Instruction template was adapted from DeepEval's codebase https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/faithfulness/template.py.
```
ntlind, line 3: repeat: it'd be helpful to start the docstrings with a one-sentence description of what each function seeks to accomplish.

api/valor_api/backend/metrics/text_generation.py
```diff
@@ -214,18 +214,16 @@ def _setup_llm_client(
     if "client" in llm_api_params and "api_url" in llm_api_params:
         raise ValueError("Cannot specify both client and api_url.")
 
-    client = llm_api_params.get("client")
-    if client is not None:
-        if client == "openai":
+    client_name = llm_api_params.get("client")
+    if client_name is not None:
+        if client_name == "openai":
```
ntlind, line 219: nit: this would be a use case for `match case` instead of multiple if statements

docs/metrics.md
```diff
@@ -310,7 +310,7 @@ print(detailed_evaluation)
 
 ## General Text Generation Metrics
 
-The general text generation metrics apply to a broad set of text generation tasks. These metrics don't compare to any groundtruths and don't require any sort of context. The metrics are evaluated purely based on the predicted text.
+The general text generation metrics apply to a broad set of text generation tasks. These metrics don't compare to any groundtruths and don't require any contexts. The metrics are evaluated purely based on the predicted text.
```
ntlind, line 313:
```suggestion
The general text generation metrics apply to a broad set of text generation tasks. These metrics don't compare to any groundtruths and don't require any additional context. The metrics are evaluated purely based on the predicted text.
```

ntlind: using the plural of "context" throughout these docs seems grammatically incorrect to me

api/valor_api/backend/core/llm_clients.py
bnativi: What is the purpose of private functions? I've seen private functions as class functions, but I am defining these instructions outside of any class. 

I could make these functions class functions. DeepEval uses for example an AnswerRelevanceTemplate class and has the instructions as class functions with the @staticmethod decorator. I could just have an LLMInstructions class and make all of the instructions class functions of that class?

api/valor_api/backend/core/llm_instructions_analysis.py
bnativi: Agreed. The edgier parts are only there because I copied them over from DeepEval. I did another pass through and removed the edgier examples.

docs/metrics.md
bnativi: I changed the one you commented on above and I also changed "contradicts the contexts" to "contradicts any context", but for the most part the use of contexts seems correct. I looked at discussion online of how to use context vs contexts and I also copy and pasted a few sections into chatgpt to ask what was correct. In those sections, it said that the use of contexts was fine, and when I changed to context, it corrected it back to contexts. 

Was there any other spot in the docs where you felt the use of contexts was incorrect?

czaloom (approved):

PR #691, opened 2024-07-30
adjusted iou computation

czaloom:
Attempting to fix intermittent slowdowns in object detection benchmark.

# Changes
- Added `.isnot(None)` clause to ensure that only annotations with the requested type are included.
- Caching results of IoU computation in a table.

### DIFF
--- a/api/valor_api/backend/metrics/detection.py
+++ b/api/valor_api/backend/metrics/detection.py
@@ -27,19 +27,6 @@ from valor_api.backend.query import generate_query, generate_select
 from valor_api.enums import AnnotationType
 
 
-def profiler(fn):
-    def wrapper(*args, **kwargs):
-        import time
-
-        print(f">>>> {fn.__name__}")
-        start = time.time()
-        result = fn(*args, **kwargs)
-        print(f"<<<< {fn.__name__} - {round(time.time() - start, 2)}")
-        return result
-
-    return wrapper
-
-
 @dataclass
 class RankedPair:
     dataset_name: str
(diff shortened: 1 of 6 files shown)


migrations/sql/00000013_iou_cache.up.sql
```diff
@@ -0,0 +1,12 @@
+create table iou
```
ntlind, line 1: is there a risk that this table will become prohibitively massive over time? the cross-join involved with DetailedPRCurve requires tons of IOU calculations

api/valor_api/backend/metrics/detection.py
```diff
@@ -969,85 +1015,13 @@ def _aggregate_data(
     return (groundtruths_cte, predictions_cte, labels)
 
 
-def _compute_detection_metrics(
+@profiler
```
ntlind, line 1018: should all the profiler references be removed?

migrations/sql/00000013_iou_cache.up.sql
ntlind: I like the idea of caching, but doing it in a table seems risky

api/valor_api/backend/metrics/detection.py
czaloom: yes, i just hadnt pushed this commit yet

migrations/sql/00000013_iou_cache.up.sql
czaloom: This is a valid concern but within a dataset-model pairing the IoU's are only computed and stored once so there is a fixed limit to the size of this table. The cross-join for detailed pr curves is still limited per datum and compared to storing a raster we are only storing a lightweight row so this should be easier on memory usage as well.

There are always going to be tradeoffs. The benefit of this solution is that subsequent evaluations (with filters or different metrics to return) will be able to reuse already computed IoUs. This will mean the first computation may take minutes but everything else will be done in seconds.

czaloom: This also separates the IoU computation from the broader metric computation which i believe was causing   smaller postgres instances to alter their normal planning.

ekorman: i also have this concern but i think its worth getting in. might want to think about mechanisms for clearing the cache if the db gets big...

czaloom: Currently, the cached IoUs are tied to the lifetime of the dataset/model.

PR #700, opened 2024-08-05
don't allow commas in model and dataset names

ekorman:
this was causing things like `get_evaluations` to fail since the query params are string separated model names

### DIFF
--- a/migrations/sql/00000013_disallow_commas.sql
+++ b/migrations/sql/00000013_disallow_commas.sql
@@ -1,0 +1,66 @@
+-- this migration goes  through dataset names and model names and replaces commas with underscores
+-- if the resulting name happens to exist, it adds underscores until it gets a name that doesn't
+-- this will update the following tables: Model, Dataset, Evaluation
+BEGIN;
+
+-- Function to get a unique name by adding underscores
+CREATE OR REPLACE FUNCTION get_unique_name(base_name TEXT, table_name TEXT)
+RETURNS TEXT AS $$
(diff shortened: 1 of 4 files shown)


ekorman:
the migration file replaces dataset and model names with commas to underscores (and adds enough underscores to guarantee uniqueness). it then updates the model, dataset, and evaluation tables accordingly.

here are screenshots of before and after running the migration command on some sample data

before:
image
image
image

after:
image
image
image

migrations/sql/00000013_disallow_commas.sql
```diff
@@ -0,0 +1,66 @@
...
+            FROM jsonb_array_elements_text(dataset_names) AS elem
+        )
+        WHERE dataset_names @> jsonb_build_array(old_name);
+    END LOOP;
+END;
+$$;
+
+DROP FUNCTION get_unique_name;
+
+COMMIT;
```
czaloom, line 66: Im not sure the `COMMIT` is required since this is a migration.

czaloom:
Does the migration run properly without `up` and `down` files?

migrations/sql/00000013_disallow_commas.sql
ekorman: i think you're right, i remvoed it

ekorman:
> Does the migration run properly without `up` and `down` files?

not sure but i added an empty `down` file

czaloom (approved):

PR #714, opened 2024-08-21
Fix Filter Docs

czaloom:
# Changes
- Added docstrings.
- Overhauled docs organization.
- Added filter-related docs.


client/valor/schemas/filters.py
```diff
@@ -41,6 +41,47 @@
 
 @dataclass
 class Filter:
+    """
+    Filter object.
```
ntlind, line 45:
```suggestion
    An object used to filter evaluations using other Valor objects (e.g., datasets, models, datums, etc.).
```

docs/client_api/filtering/Conditions/Gte.md
```diff
@@ -0,0 +1,16 @@
+::: valor.schemas.Gte
```
ntlind, line 1: could all of the "Conditions" exist on one page, and all of the "Logic" on another, rather than having a separate page for each individual class? in this setup, it seems like it would take a lot of clicking between pages for a user to figure out what to use. some of the docs for these classes (like most of the "Conditions") are also very barebones, so it seems odd to break them out separately

czaloom: image

I think this leads to a confusing docs layout. Do you have any suggestions on how to improve this?

client/valor/schemas/filters.py
czaloom: Updated with `A data class that encapsulates filter conditions for various Valor components.`

docs/client_api/filtering/Conditions/Gte.md
ntlind: I don't think that format is visually confusing, but it's weird that you're mixing operators (Gt, Gte, etc.) with other stuff (Condition, Function) under a header called "Operators". maybe move these latter two docstrings into Filter or another category?

ntlind (approved):

PR #724, opened 2024-08-27
Refactor object detection in `valor_core` to improve speed

ntlind:
## Improvements
- Add back a faster IOU calculation for axis-aligned bboxes
- Add [`pandas[performance]` as a dependency](https://pandas.pydata.org/docs/getting_started/install.html#performance-dependencies-recommended). Use `@numba.jit` to speed-up IOU calculations
- Replace `apply` and `lambda` calls where possible to reduce runtime
- Move an expected output fixture from `_inputs` to `_outputs`

## Performance

Base OD evaluation for `AnnotationType.Box` is down to ~11 seconds on the GH runner (previously ~21), and ~6 seconds locally


core/tests/conftest_inputs.py
```diff
@@ -3046,243 +3046,6 @@ def mammal_label_map():
     }
 
 
-@pytest.fixture
```
czaloom, line 3049: Why did this input get removed?

czaloom (approved):

PR #726, opened 2024-08-28
Fix OD benchmarks and disallow calculating DetailedPRCurves when using AnnotationType.RASTER

ntlind:
## Improvements
- Fix a bug in the benchmarking script where we weren't actually telling `valor_core` to calculate `DetailedPrecisionRecallCurve`
- Throw a `NotImplementedError` when a user tries to calculate `DetailedPrecisionRecallCurve` using `AnnotationType.RASTER`.
    - Currently, `valor_core` will try to produce (datum_uid, *mask*) pairs, which easily leads to OOM issues since we're creating sets of tuples, each containing relatively large masks
    -  These curves were never intended to be used with rasters (just bounding boxes and polygons), so I'm recommending we just throw a `NotImplementedError` for now. when we re-prioritize DetailedPRCurves in the future, we can discuss what the right output is for this metric.
- Disable `compute_detailed` for the Box and Polygon OD benchmarks since a) they make the tests take a while to run (~80 seconds for just one run of 5000 `Box` datums; 213 seconds for one Polygon run) and b) profiling and improving these PR curves is low priority right now. We can try adding them back after Justin merges his PRs.
- Update the `valor_core` OD benchmarks to use the ValorDetectionManager class 
- Refactor the raster IOU function to make it easier to read

### DIFF
--- a/core/benchmarks/object-detection/benchmark_script.py
+++ b/core/benchmarks/object-detection/benchmark_script.py
@@ -466,6 +466,7 @@ if __name__ == "__main__":
             (AnnotationType.BOX, AnnotationType.BOX),
         ],
         limits_to_test=[5000, 5000],
+        compute_detailed=False,
     )
 
     # run polygon benchmark
@@ -474,6 +475,7 @@ if __name__ == "__main__":
             (AnnotationType.POLYGON, AnnotationType.POLYGON),
         ],
         limits_to_test=[5000, 5000],
+        compute_detailed=False,
     )
     # run raster benchmark
     run_benchmarking_analysis(
--- a/core/benchmarks/object-detection/benchmark_script.py
+++ b/core/benchmarks/object-detection/benchmark_script.py
@@ -21,6 +21,7 @@ from valor_core import (
     Polygon,
     Prediction,
     Raster,
+    ValorDetectionManager,
     enums,
     evaluate_detection,
 )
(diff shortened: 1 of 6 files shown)


core/benchmarks/object-detection/benchmark_script.py
```diff
@@ -421,12 +475,13 @@ def run_benchmarking_analysis(
...
+        compute_detailed=False,
     )
-
     # run raster benchmark
     run_benchmarking_analysis(
         combinations=[
             (AnnotationType.RASTER, AnnotationType.RASTER),
         ],
-        limits_to_test=[500, 500],
+        limits_to_test=[500],
```
czaloom, line 485: Can we switch back to two runs?

core/valor_core/detection.py
```diff
@@ -643,6 +621,21 @@ def _calculate_detailed_pr_metrics(
     ) or (detailed_pr_joint_df is None):
         return []
 
+    if _check_if_series_contains_masks(
+        detailed_pr_joint_df.loc[
+            detailed_pr_joint_df["converted_geometry_pd"].notnull(),
```
czaloom, line 626: Should this be a gt check it looks like this statement would result in `pd or pd`

core/benchmarks/object-detection/benchmark_script.py
ntlind: thank you! yes, should have changed that back

core/valor_core/geometry.py
```diff
@@ -244,3 +245,42 @@ def is_rotated(bbox: list[tuple[float, float]]) -> bool:
...
+
+    Returns
+    ----------
+    pd.Series
+        A Series indicating the union of two masks.
+    """
+
+    return np.sum(row["converted_geometry_gt"]) + np.sum(
+        row["converted_geometry_pd"]
+    )
```
czaloom, line 286: I understand this is a component of the union operation, but this function is performing the sum of pixels across two images not a union.

The union could be computed either as sum(A) + sum(B) - sum(and(A, B))

or 

```
np.logical_or(
        row["converted_geometry_pd"], row["converted_geometry_gt"]
    ).sum()
```

ntlind: I see what you mean. I'm changed this function name to be more clear; I'd rather keep the intersection series out of this function since I think it's a it easier to follow

czaloom (approved):

PR #734, opened 2024-09-04
WIP Optimized Detection Metics

jqu-striveworks:
vroom

### DIFF
--- a/core/valor_core/detection2.py
+++ b/core/valor_core/detection2.py
@@ -1,0 +1,497 @@
+import time
+
+import numpy as np
+import pandas as pd
+from valor_core import enums, metrics, schemas
+
+
+def bbox_to_useful(bbox):
+    coords = bbox.value[0]
+    x_values = {c[0] for c in coords[:3]}
+    y_values = {c[1] for c in coords[:3]}
+    x_min = min(x_values)
+    x_max = max(x_values)
+    y_min = min(y_values)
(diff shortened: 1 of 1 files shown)


jqu-striveworks:
Pretty confident that it works and should match Valor core metrics. EXCEPT for small change.

The AP metrics will be slightly different since I opted not to use the 101-point interpolation. Instead, I calculate the exact area under the interpolated curve, rather than Avg. 101 samples from it. It's more accurate and faster to calculate so ¯\_(ツ)_/¯ why not i guess.

Interpolated curve shown [here](https://learnopencv.com/wp-content/uploads/2022/08/mean-average-precision-map-final-precision-interpolated-graph.png) in purple.

ekorman:
> Pretty confident that it works and should match Valor core metrics. EXCEPT for small change.
> 
> The AP metrics will be slightly different since I opted not to use the 101-point interpolation. Instead, I calculate the exact area under the interpolated curve, rather than Avg. 101 samples from it. It's more accurate and faster to calculate so ¯_(ツ)_/¯ why not i guess.
> 
> Interpolated curve shown [here](https://learnopencv.com/wp-content/uploads/2022/08/mean-average-precision-map-final-precision-interpolated-graph.png) in purple.

not sure what you mean by "more accurate"-- using 101-point interpolation is the standard definition of the metric.

PR #747, opened 2024-09-13
Add Text Generation Metrics to Valor Core

bnativi:


core/valor_core/managers.py
```diff
@@ -590,3 +598,229 @@ def evaluate(self):
...
+            prediction_text = m["parameters"].get("prediction", None)
+            prediction_context_list = m["parameters"].get("context_list", None)
+
+            # Set the metric value of the correct row.
+            conditional_series = self.joint_df.apply(
+                conditions,
+                axis=1,
+                args=(datum_uid, prediction_text, prediction_context_list),
+            )
+            self.joint_df.loc[conditional_series, metric_name] = value
```
bnativi, line 809: None of the code in the streaming manager prevents the addition of duplicate predictions. If this function is called with a prediction annotation that matches the exact same Datum, prediction.text and prediction.context_list, then even though that data is already in the joint_df, the metrics will be computed again and a new row will be added to the joint_df. Further, because of the way that I add the metric results to the joint_df, the second metric results will also overwrite the first metric results for the first row in the joint_df. This is potentially an issue that an old row in the dataframe could get updated, however this is not a high priority issue. The most likely best fix for this issue would be to prevent the addition of identical prediction annotations, but that would take some work to setup and test.

core/valor_core/text_generation.py
```diff
@@ -0,0 +1,634 @@
...
+    utilities.validate_metric_parameters(
+        metrics_to_return=metrics_to_return,
+        metric_params=metric_params,
+    )
+
+    unique_datum_counts = len(set([p.datum.uid for p in predictions]))
+
+    # Generate a list of data tuples, where each prediction is matched with its corresponding datum and the associated groundtruths.
+    # TODO consider changing to dataframe
+    data = []
```
bnativi, line 594: I was considering using a pandas dataframe like the other task types use. However I don't think this code needs a dataframe for any reason (I'm not doing any computations that would benefit from having a dataframe), so I don't think this change would be worth it for now.

core/valor_core/utilities.py
```diff
@@ -534,56 +573,84 @@ def _convert_groundtruth_or_prediction_to_dataframe(
...
+                                "label_key": label_key,
+                                "label_value": label_value,
+                                "label_id": label_id,
+                                "id": id_,
+                            }
+                        )
+            else:
                 id_ = (
-                    str(ann_id) + str(i) + str(j) + str(k)
+                    str(ann_id) + str(i) + str(j) + "0"
```
bnativi, line 635: I don't have any testing for this id_ because none of my tests actually use this dataframe structure for the data.

bnativi: I'm fine to remove all the text gen stuff from these utilities if that's what you guys think is best, as some of those additions are untested as I am not using the same dataframe structure

ekorman (approved):

PR #749, opened 2024-09-16
Lite Classification

czaloom:
# Additions
- Added classification to the `valor-lite` package.


lite/valor_lite/classification/manager.py
```diff
@@ -0,0 +1,651 @@
...
+                )
+
+            gt_keys = set(keyed_groundtruths.keys())
+            pd_keys = set(keyed_predictions.keys())
+            joint_keys = gt_keys.intersection(pd_keys)
+
+            gt_unique_keys = gt_keys - pd_keys
+            pd_unique_keys = pd_keys - gt_keys
+            if gt_unique_keys or pd_unique_keys:
+                raise ValueError()
```
ntlind, line 488: add "label keys must match" error message

lite/valor_lite/classification/annotation.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
```
ntlind, line 1: would it make sense to rename this file to something like "classification_schemas"? "annotation" seems like an odd choice

lite/valor_lite/classification/computation.py
```diff
@@ -0,0 +1,330 @@
+import numpy as np
+from numpy.typing import NDArray
+
+# 0 gt label index
```
ntlind, line 4: remove these comments since you added this description to the docstring?

lite/valor_lite/classification/computation.py
```diff
@@ -0,0 +1,330 @@
...
+    n_labels = label_metadata.shape[0]
+    n_label_keys = np.unique(label_metadata[:, 2]).size
+    n_scores = score_thresholds.shape[0]
+
+    pd_labels = data[:, 1].astype(int)
+
+    mask_matching_labels = np.isclose(data[:, 0], data[:, 1])
+    mask_score_nonzero = ~np.isclose(data[:, 2], 0.0)
+
+    # ==== calculate ROCAUC ====
```
ntlind, line 68: instead of adding these comments into one long function, could you split this function's contents into smaller functions (e.g., `_calculate_rocauc`)?

lite/valor_lite/classification/computation.py
```diff
@@ -0,0 +1,330 @@
...
+        mask_pds = pd_labels == label_idx
+
+        true_positives[label_idx] = mask_matching_labels[mask_pds]
+        false_positives[label_idx] = ~mask_matching_labels[mask_pds]
+        scores[label_idx] = data[mask_pds, 2]
+
+    cumulative_fp = np.cumsum(false_positives, axis=1)
+    cumulative_tp = np.cumsum(true_positives, axis=1)
+
+    fpr = np.zeros_like(true_positives, dtype=np.float64)
```
ntlind, line 99: what does "tpr" and "fpr" stand for? could these names be more descriptive?

ntlind, line 99: what does tpr and fpr stand for? rolling true positives / false positives? if so, could you make these names more descriptive?

lite/valor_lite/classification/computation.py
```diff
@@ -0,0 +1,330 @@
...
+    )
+
+
+def compute_detailed_counts(
+    data: NDArray[np.floating],
+    label_metadata: NDArray[np.int32],
+    score_thresholds: NDArray[np.floating],
+    n_samples: int,
+) -> NDArray[np.floating]:
+    """
```
ntlind, line 207: fix docstring?

lite/valor_lite/classification/manager.py
```diff
@@ -0,0 +1,651 @@
...
+    @property
+    def ignored_prediction_labels(self) -> list[tuple[str, str]]:
+        glabels = set(np.where(self._label_metadata[:, 0] > 0)[0])
+        plabels = set(np.where(self._label_metadata[:, 1] > 0)[0])
+        return [
+            self.index_to_label[label_id] for label_id in (plabels - glabels)
+        ]
+
+    @property
+    def missing_prediction_labels(self) -> list[tuple[str, str]]:
```
ntlind, line 88: add docstrings to all of these properties to describe how they're calculated and why they're useful?

lite/valor_lite/classification/manager.py
```diff
@@ -0,0 +1,651 @@
...
+
+        return Filter(
+            indices=np.where(mask_pairs)[0],
+            label_metadata=label_metadata,
+            n_datums=n_datums,
+        )
+
+    def evaluate(
+        self,
+        score_thresholds: list[float] = [0.5],
```
ntlind, line 209: add full docstring?

lite/valor_lite/classification/manager.py
```diff
@@ -0,0 +1,651 @@
...
+                    value=f1_score[:, label_idx].tolist(),
+                    **kwargs,
+                )
+            )
+
+        return metrics
+
+    def compute_detailed_counts(
+        self,
+        score_thresholds: list[float] = [
```
ntlind, line 312: add docstring?

lite/valor_lite/classification/manager.py
```diff
@@ -0,0 +1,651 @@
...
+                    for score_idx in range(n_scores)
+                ],
+                scores=score_thresholds,
+                label=self.index_to_label[label_idx],
+            )
+            for label_idx in range(n_labels)
+        ]
+
+
+class DataLoader:
```
ntlind, line 410: add docstring?

lite/valor_lite/classification/annotation.py
czaloom: My reasoning is as follows:

a classification is a form of annotation that applies labels and scores to a datum.

a detection is a form of annotation that applies labeled + scored geometries to a datum

lite/valor_lite/classification/computation.py
czaloom: True-Positive and False-Positive Rate. These are standard names for the variables.

lite/tests/classification/test_dataloader.py
```diff
@@ -0,0 +1,44 @@
...
+    assert loader._evaluator.n_datums == 1
+
+
+def test_mismatch_label_keys(
+    classifications_no_groundtruths: list[Classification],
+    classifications_no_predictions: list[Classification],
+):
+    loader = DataLoader()
+
+    with pytest.raises(ValueError):
```
ntlind, line 40: when looking into benchmarks for object detection: we allow the user to pass an empty list of annotations in a groundtruth/prediction without throwing an error. we also allow the user to not pass any groundtruth or prediction annotations at all without throwing an error (mentioned this on slack yesterday, and you said this is likely in real world scenarios).

the tests here seem to show that we don't want to allow the user to pass empty lists of groundtruth / prediction annotations into valor. are we inconsistent between classification / detection, or between add_data / add_data_from_valor_dict for detections?

lite/tests/classification/test_dataloader.py
```diff
@@ -0,0 +1,44 @@
...
+        loader.add_data(classifications_no_groundtruths)
+
+    with pytest.raises(ValueError):
+        loader.add_data(classifications_no_predictions)
```
ntlind, line 44: in `test_evaluate_classification_model_with_no_predictions`, we show that we want users to be able to pass an empty set of predictions and still get back classification metrics (with zeroes for all values). doesn't seem like this test was added, and this behavior may be disallowed as shown by this line

lite/tests/classification/test_detailed_counts.py
```diff
@@ -0,0 +1,785 @@
...
+            "type": "DetailedCounts",
+        },
+    ]
+    for m in actual_metrics:
+        assert m in expected_metrics
+    for m in expected_metrics:
+        assert m in actual_metrics
+
+
+def test_detailed_counts_mutliclass(
```
ntlind, line 413: typo

ntlind (changes requested):

lite/tests/classification/test_counts.py
```diff
@@ -0,0 +1,862 @@
...
+            },
+        },
+        {
+            "type": "Counts",
+            "value": {
+                "tp": [
+                    0,
+                    0,
+                    0,
+                ],  # TODO shouldn't the sixth groundtruth be a tp at low score thresholds? the groundtruth is "dog", and both "cat" and "dog" share a score of .4
```
ntlind, line 288: I expanded this test a bit to include more score thresholds and added comments in the form of "TODO"s

lite/tests/classification/test_dataloader.py
ntlind: update: we decided on 9/30 to throw an error if the use tries to pass empty groundtruth or prediction sets

lite/tests/classification/test_rocauc.py
```diff
@@ -0,0 +1,254 @@
...
+            "value": 0.43125,
+            "parameters": {
+                "label_key": "color",
+            },
+        },
+    ]
+    for m in actual_metrics:
+        assert m in expected_metrics
+    for m in expected_metrics:
+        assert m in actual_metrics
```
ntlind, line 254: I found an unexpected result when writing this test:

```python

def test_nonoverlapping_labels():
    animal_gts = ["cat"]
    animal_pds = [
        {"bird": 0.6, "dog": 0.2, "cat": 0.2},
    ]

    classifications = [
        Classification(
            uid=f"uid{idx}",
            groundtruths=[("animal", gt)],
            predictions=[("animal", key) for key, _ in pd.items()],
            scores=[score for score in pd.values()],
        )
        for idx, (gt, pd) in enumerate(zip(animal_gts, animal_pds))
    ]

    loader = DataLoader()
    loader.add_data(classifications)
    evaluator = loader.finalize()

    metrics = evaluator.evaluate(score_thresholds=[0.05, 0.5, 0.95])
    metrics_to_test = [
        MetricType.Accuracy,
        MetricType.Precision,
        MetricType.ROCAUC,
    ]

    actual_metrics = [
        m.to_dict() for key in metrics_to_test for m in metrics[key]
    ]
    expected_metrics = [
        {
            "type": "Accuracy",
            "value": [0.0, 0.0, 0.0],
            "parameters": {
                "score_thresholds": [0.05, 0.5, 0.95],
                "hardmax": True,
                "label": {"key": "animal", "value": "cat"},
            },
        },
        {
            "type": "Precision",
            "value": [0.0, 0.0, 0.0],
            "parameters": {
                "score_thresholds": [0.05, 0.5, 0.95],
                "hardmax": True,
                "label": {"key": "animal", "value": "cat"},
            },
        },
        {
            "type": "ROCAUC",
            "value": 0,
            "parameters": {"label": {"key": "animal", "value": "cat"}},
        },
        # TODO should we return ROCAUC == 0 for "bird" and "dog" here if we don't return any Precision / Accuracy metrics for these two labels?
        {
            "type": "ROCAUC",
            "value": 0,
            "parameters": {"label": {"key": "animal", "value": "bird"}},
        },
        {
            "type": "ROCAUC",
            "value": 0,
            "parameters": {"label": {"key": "animal", "value": "dog"}},
        },
    ]
    for m in actual_metrics:
        assert m in expected_metrics
    for m in expected_metrics:
        assert m in actual_metrics
```

should we return ROCAUC metrics for `bird` and `dog` in this example when we don't return Accuracy or Precision metrics for these two labels? seems like we shouldn't return any metrics for "bird" or "dog" in this case

lite/tests/classification/test_counts.py
czaloom: The existing results are correct for the input args. For the regular Counts metric in this test we are taking the hardmax. This means that one prediction "wins" and the others are considered to be false.

lite/tests/classification/test_rocauc.py
czaloom: Adjusted the evaluate call to filter out ROCAUC's for labels that are a part of `ignored_prediction_labels`

czaloom: This is what Accuracy, F1, Recall and Precision were already doing

lite/tests/classification/test_dataloader.py
```diff
@@ -0,0 +1,44 @@
...
+    assert loader._evaluator._detailed_pairs.shape == (10, 5)
+
+    assert set(loader._evaluator.label_key_to_index.keys()) == {
+        "class_label",
+    }
+    assert len(loader._evaluator.index_to_label) == 11
+    assert loader._evaluator.n_datums == 1
+
+
+def test_mismatch_label_keys(
```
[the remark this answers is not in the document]
czaloom: Added to the test!

lite/valor_lite/classification/manager.py
```diff
@@ -0,0 +1,854 @@
...
+                score_thresholds=score_thresholds,
+                hardmax=hardmax,
+                number_of_examples=number_of_examples,
+            )
+
+        for metric in set(metrics.keys()):
+            if metric not in metrics_to_return:
+                del metrics[metric]
+
+        return metrics
```
ntlind, line 356: can you return an array of dictionaries for each metric key, instead of a dictionary of metric objects? that would help resolve this ticket: https://github.com/orgs/Striveworks/projects/2/views/1?pane=issue&itemId=80550739

lite/valor_lite/classification/manager.py
```diff
@@ -0,0 +1,854 @@
...
+        Ground truth labels that are not present in the prediction set.
+        """
+        glabels = set(np.where(self._label_metadata[:, 0] > 0)[0])
+        plabels = set(np.where(self._label_metadata[:, 1] > 0)[0])
+        return [
+            self.index_to_label[label_id] for label_id in (glabels - plabels)
+        ]
+
+    @property
+    def metadata(self) -> dict:
```
ntlind, line 106: add Returns to this docstring to make consistent with other docstrings in this file?

lite/valor_lite/classification/manager.py
```diff
@@ -0,0 +1,854 @@
...
+        self.label_key_to_index: dict[str, int] = dict()
+        self.label_index_to_label_key_index: dict[int, int] = dict()
+
+        # computation caches
+        self._detailed_pairs = np.array([])
+        self._label_metadata = np.array([], dtype=np.int32)
+        self._label_metadata_per_datum = np.array([], dtype=np.int32)
+
+    @property
+    def ignored_prediction_labels(self) -> list[tuple[str, str]]:
```
ntlind, line 84: add Returns to this docstring to make consistent with other docstrings in this file?

lite/tests/classification/test_counts.py
czaloom: There is no winner here b/c the score_threshold is `0.5` and the highest score is `0.4`

lite/tests/classification/test_counts.py
```diff
@@ -0,0 +1,857 @@
...
+
+def test_counts_with_example(
+    classifications_two_categeories: list[Classification],
+):
+
+    loader = DataLoader()
+    loader.add_data(classifications_two_categeories)
+    evaluator = loader.finalize()
+
+    metrics = evaluator.evaluate(score_thresholds=[0.5, 0.95])
```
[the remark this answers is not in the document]
czaloom: There is no point as they produce the same results as 0.5...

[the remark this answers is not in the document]
czaloom: Its just cluttering up the test. If you review the data this test covers its clear that the results will be the same between 0.05 and 0.5.

lite/valor_lite/classification/computation.py
```diff
@@ -0,0 +1,411 @@
...
+
+    n_labels = label_metadata.shape[0]
+    n_label_keys = np.unique(label_metadata[:, 2]).size
+    n_scores = score_thresholds.shape[0]
+
+    pd_labels = data[:, 2].astype(int)
+
+    mask_matching_labels = np.isclose(data[:, 1], data[:, 2])
+    mask_score_nonzero = ~np.isclose(data[:, 3], 0.0)
+    mask_hardmax = data[:, 4] > 0.5
```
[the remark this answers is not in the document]
czaloom: A hard max is either 0.0 or 1.0 so this is just checking for that.

The hardmax's are computed using np.argmax in the `add_data` method.

lite/valor_lite/classification/manager.py
czaloom: Im not sure what you mean in this ticket, but the output currently matches obj detection so im going to consider this suggestion out of scope for this PR.

lite/valor_lite/classification/manager.py
czaloom: Since its a property the one-liner describes the output.

It would be kinda redundant to say:

```
"""
Evaluation metadata.

Returns
---------
dict
    Evaluation metadat

```

lite/tests/classification/test_counts.py
ntlind: yes, I asked you to include the score threshold of 0.05 to check for this (which I had added, and you later removed)

lite/valor_lite/classification/manager.py
ntlind: we'd discussed not returning metric objects to the user since they'd just have to convert them to dictionaries, anyway. not considering this OOS

lite/tests/classification/test_counts.py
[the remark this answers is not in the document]
czaloom: We are choosing to keep the argmax implementation per a conversation with Eric.

lite/valor_lite/classification/manager.py
czaloom: Added the `as_dict` arg to the `evaluate` call to provide the requested behavior.

lite/tests/classification/test_counts.py
czaloom: This is included now!

lite/tests/classification/test_accuracy.py
```diff
@@ -84,10 +84,10 @@ def test_accuracy_basic(basic_classifications: list[Classification]):
         "missing_prediction_labels": [],
     }
 
-    metrics = evaluator.evaluate(score_thresholds=[0.25, 0.75])
+    metrics = evaluator.evaluate(score_thresholds=[0.25, 0.75], as_dict=True)
 
     # test Accuracy
-    actual_metrics = [m.to_dict() for m in metrics[MetricType.Accuracy]]
+    actual_metrics = [m for m in metrics[MetricType.Accuracy]]
```
ntlind, line 90: nit: it's probably cleaner to pass Accuracy to `metrics_to_return`, then you don't need this step at all

ntlind (approved):

lite/tests/classification/test_accuracy.py
czaloom: The alt would be 

```
metrics = evaluate.evaluate(
    ...,
    metrics_to_return=[MetricType.Accuracy],
)[MetricType.Accuracy]

actual_metrics = [m for m in metrics]
```

which im not a huge fan of visually

PR #751, opened 2024-09-19
Fix Detailed Counts in Valor-Lite

czaloom:
# Changes
- Renamed threshold parameters to be more explicit.
- Renamed `DetailedPrecisionRecallCurve` to `DetailedCounts` since the metric contains more than just precision and recall.
- Fixed computation for detailed counts. FN were not being tracked properly.


ntlind:
can you add the missing parameters from the API and `core` as part of this PR (e.g., pr_curve_iou_threshold, pr_curve_max_examples, etc.)?

ntlind (approved):

PR #759, opened 2024-09-25
Enable Bitmasks and Polygons in `lite`

ntlind:
## Improvements
- Enable polygons (e.g., rotated bounding boxes) and bitmasks in `lite`
- Add additional tests from `core` in `lite`
- Confirmed that coverage and benchmark times didn't change


lite/tests/detection/conftest.py
```diff
@@ -309,9 +403,9 @@ def false_negatives_single_datum_detections() -> list[Detection]:
 
 
 @pytest.fixture
-def false_negatives_two_datums_one_empty_low_confidence_of_fp_detections() -> list[
-    Detection
-]:
+def false_negatives_two_datums_one_empty_low_confidence_of_fp_detections() -> (
+    list[Detection]
+):
```
czaloom, line 408: Why make this change? It seems like this is done across this file.

lite/tests/detection/test_average_precision.py
czaloom, line 1: Instead of making tests really long, i would prefer you spin off functions like `test_ap_metrics_for_rotated_bboxes`.

When running tests its more clear what is actually being tested.

lite/tests/detection/conftest.py
ntlind: this fixture returns a list of Detection, not just one Detection. it was throwing a type error in my linter

lite/tests/detection/test_average_precision.py
```diff
@@ -623,161 +631,175 @@ def test_ap_false_negatives_two_datums_one_only_with_different_class_high_confid
...
-        "n_groundtruths": 3,
-        "n_labels": 4,
-        "n_predictions": 4,
-    }
+def test_ap_ranked_pair_ordering(
+    detection_ranked_pair_ordering: Detection,
+    detection_ranked_pair_ordering_with_bitmasks: Detection,
+    detection_ranked_pair_ordering_with_polygons: Detection,
+):
+    for input_, method in [
```
ntlind, line 639: all of these tests are the same as before, I'm just iterating through multiple inputs and methods (e.g., `add_bounding_boxes`) to make sure they all return the same output

lite/tests/detection/conftest.py
czaloom: It looks like it and the others were already defined as lists. Why the change from `-> list[Detection]` to `-> (list[Detection])`

lite/tests/detection/conftest.py
```diff
@@ -354,9 +448,9 @@ def false_negatives_two_datums_one_empty_low_confidence_of_fp_detections() -> li
 
 
 @pytest.fixture
-def false_negatives_two_datums_one_empty_high_confidence_of_fp_detections() -> list[
-    Detection
-]:
+def false_negatives_two_datums_one_empty_high_confidence_of_fp_detections() -> (
+    list[Detection]
+):
```
czaloom, line 453: Another instance of the above ^

lite/tests/detection/conftest.py
```diff
@@ -399,9 +493,9 @@ def false_negatives_two_datums_one_empty_high_confidence_of_fp_detections() -> l
 
 
 @pytest.fixture
-def false_negatives_two_datums_one_only_with_different_class_low_confidence_of_fp_detections() -> list[
-    Detection
-]:
+def false_negatives_two_datums_one_only_with_different_class_low_confidence_of_fp_detections() -> (
+    list[Detection]
+):
```
czaloom, line 498: Another instance of the above ^

lite/tests/detection/conftest.py
```diff
@@ -452,9 +546,9 @@ def false_negatives_two_datums_one_only_with_different_class_low_confidence_of_f
 
 
 @pytest.fixture
-def false_negatives_two_images_one_only_with_different_class_high_confidence_of_fp_detections() -> list[
-    Detection
-]:
+def false_negatives_two_images_one_only_with_different_class_high_confidence_of_fp_detections() -> (
+    list[Detection]
+):
```
czaloom, line 551: Another instance of the above ^

lite/valor_lite/detection/manager.py
```diff
@@ -51,6 +57,74 @@
...
 
 
+def _get_annotation_key(
+    annotation_type: type[BoundingBox] | type[Polygon] | type[Bitmask],
+) -> str:
+    """Get the correct JSON key to extract a given annotation type."""
+
+    if issubclass(annotation_type, BoundingBox):
+        return "bounding_box"
+    if issubclass(annotation_type, Polygon):
```
czaloom, line 67: Why use `issubclass` rather than `isinstance`?

lite/valor_lite/detection/manager.py
```diff
@@ -51,6 +57,74 @@
 """
 
 
+def _get_annotation_key(
```
czaloom, line 60: nit: could you rename this to `_get_valor_dict_annotation_key` since its only used in valor dict ingestion?

lite/valor_lite/detection/manager.py
```diff
@@ -817,45 +908,129 @@ def add_data(
...
-                ] = np.array(gann.extrema)
+                if not isinstance(gann, annotation_type):
+                    raise ValueError(
+                        f"Expected {annotation_type}, but annotation is of type {type(gann)}."
+                    )
+
+                if annotation_type == BoundingBox:
+                    self._evaluator.groundtruth_examples[uid_index][
+                        gidx
+                    ] = _get_annotation_representation(obj=gann)
```
czaloom, line 921: Could you add in a bbox conversion step for polygon and bitmask (e.g. `Bitmask.to_box()`) so that we can cache bounding box examples for those annotations too?

lite/valor_lite/detection/manager.py
```diff
@@ -817,45 +908,129 @@ def add_data(
...
+
                 for glabel in gann.labels:
                     label_idx, label_key_idx = self._add_label(glabel)
                     self.groundtruth_count[label_idx][uid_index] += 1
+                    representation = _get_annotation_representation(obj=gann)
```
czaloom, line 926: This should be moved out of this loop as it is just performing the same op for all labels.

lite/valor_lite/detection/manager.py
```diff
@@ -817,45 +908,129 @@ def add_data(
...
-                ] = np.array(pann.extrema)
+                if not isinstance(pann, annotation_type):
+                    raise ValueError(
+                        f"Expected {annotation_type}, but annotation is of type {type(pann)}."
+                    )
+
+                if annotation_type == BoundingBox:
+                    self._evaluator.prediction_examples[uid_index][
+                        pidx
+                    ] = _get_annotation_representation(obj=pann)
```
czaloom, line 944: Same as above, can we cache box examples of polygon and bitmask annotations?

lite/valor_lite/detection/manager.py
```diff
@@ -817,45 +908,129 @@ def add_data(
...
                 for plabel, pscore in zip(pann.labels, pann.scores):
                     label_idx, label_key_idx = self._add_label(plabel)
                     self.prediction_count[label_idx][uid_index] += 1
+                    representation = _get_annotation_representation(
```
czaloom, line 948: Same as above, can you move this out of the label loop since it only needs to be run once per annotation?

lite/valor_lite/detection/manager.py
czaloom: The examples call can share the results of a single call too.

lite/valor_lite/detection/computation.py
```diff
@@ -69,6 +68,102 @@ def compute_iou(data: NDArray[np.floating]) -> NDArray[np.floating]:
...
+            for i, (poly1, poly2) in enumerate(data)
+        ]
+    )
+
+    return intersection_ / union_
+
+
+def compute_iou(
+    data: NDArray[np.floating],
+    annotation_type: type[BoundingBox] | type[Polygon] | type[Bitmask],
```
czaloom, line 140: Since this is a routing function could you move it to the manager?

The goal here is to have the computations file be numpy in -> numpy out.

lite/benchmarks/benchmark_objdet.py
czaloom, line 1: Could you make a ticket to add benchmarking of the other annotation types?

lite/valor_lite/detection/manager.py
ntlind: made this change, but note that _add_data now returns examples for polygons / bitmasks while _add_data_from_valor_dict doesn't

lite/valor_lite/detection/computation.py
```diff
@@ -69,6 +67,73 @@ def compute_iou(data: NDArray[np.floating]) -> NDArray[np.floating]:
...
+        Computed IoU's.
+    """
+    intersection_ = np.array([np.logical_and(x, y).sum() for x, y in data])
+    union_ = np.array([np.logical_or(x, y).sum() for x, y in data])
+
+    return intersection_ / union_
+
+
+def compute_polygon_iou(
+    data: np.ndarray,
```
czaloom, line 100: nit: can you update this to `NDArray[np.floating]`

czaloom (approved):

lite/valor_lite/detection/computation.py
```diff
@@ -226,10 +226,10 @@ def compute_ranked_pairs(
 
 
 def compute_metrics(
-    data: np.ndarray,
-    label_metadata: np.ndarray,
-    iou_thresholds: np.ndarray,
-    score_thresholds: np.ndarray,
+    data: NDArray[np.floating],
+    label_metadata: NDArray[np.floating],
```
czaloom, line 230: this should be NDArray[np.int32]

PR #760, opened 2024-09-25
Added Obj Det tests to Lite

czaloom:


lite/tests/detection/test_detailed_counts.py
```diff
@@ -1770,6 +1770,9 @@ def test_detailed_counts_using_torch_metrics_example(
     ]
 
     for m in actual_metrics:
+        import json
```
ntlind, line 1773: remove print

ntlind (approved):

PR #773, opened 2024-10-02
Lite Semantic Segmentation

czaloom:
Implements Semantic Segmentation evaluation in Valor Lite.

# Metrics
- Precision
- Recall
- Accuracy
- F1
- IoU
- mIoU
- ConfusionMatrix

### IoU Confusion Matrix

```
{
    "type": "ConfusionMatrix",
    "value": {
        "confusion_matrix": {
            "v1": {
                "v1": {
                    "iou": 5000 / (10000 + 10000 - 5000)
                },  # 50% overlap
                "v2": {"iou": 0.0},
            },
            "v2": {
                "v1": {"iou": 0.0},
                "v2": {
                    "iou": 1 / (14999 + 4999 + 1)  # overlaps 1 pixel
                },
            },
        },
        "hallucinations": {
            "v1": {"percent": 5000 / 10000},  # 50% overlap
            "v2": {
                "percent": 4999 / 5000
            },  # overlaps 1 pixel out of 5000 predictions
        },
        "missing_predictions": {
            "v1": {"percent": 5000 / 10000},
            "v2": {
                "percent": 14999 / 15000
            },  # overlaps 1 pixel out of 15,000 groundtruths
        },
    },
    "parameters": {},
}
```

### DIFF
--- a/lite/tests/segmentation/test_accuracy.py
+++ b/lite/tests/segmentation/test_accuracy.py
@@ -1,0 +1,30 @@
+from valor_lite.segmentation import DataLoader, MetricType, Segmentation
+
+
+def test_accuracy_basic_segmenations(basic_segmentations: list[Segmentation]):
+    loader = DataLoader()
+    loader.add_data(basic_segmentations)
+    evaluator = loader.finalize()
+
+    metrics = evaluator.evaluate(
+        score_thresholds=[0.4, 0.6],
+        as_dict=True,
+    )
+
+    actual_metrics = [m for m in metrics[MetricType.Accuracy]]
+    expected_metrics = [
+        {
+            "type": "Accuracy",
+            "value": [
+                0.75,
+                1.0,
+            ],
+            "parameters": {
+                "score_thresholds": [0.4, 0.6],
+            },
+        },
+    ]
+    for m in actual_metrics:
+        assert m in expected_metrics
+    for m in expected_metrics:
+        assert m in actual_metrics
--- a/lite/tests/segmentation/test_accuracy.py
+++ b/lite/tests/segmentation/test_accuracy.py
@@ -6,22 +6,14 @@ def test_accuracy_basic_segmenations(basic_segmentations: list[Segmentation]):
     loader.add_data(basic_segmentations)
     evaluator = loader.finalize()
 
-    metrics = evaluator.evaluate(
-        score_thresholds=[0.4, 0.6],
-        as_dict=True,
-    )
+    metrics = evaluator.evaluate(as_dict=True)
 
     actual_metrics = [m for m in metrics[MetricType.Accuracy]]
     expected_metrics = [
         {
             "type": "Accuracy",
-            "value": [
-                0.75,
-                1.0,
-            ],
-            "parameters": {
-                "score_thresholds": [0.4, 0.6],
-            },
+            "value": 2 / 3,
+            "parameters": {},
         },
     ]
     for m in actual_metrics:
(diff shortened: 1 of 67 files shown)


lite/tests/segmentation/test_dataloader.py
```diff
@@ -0,0 +1,8 @@
+import pytest
```
ntlind, line 1: can you add a test where you check that we get all the metrics with 0 values when you pass in a groundtruth and no predictions (see `test_evaluate_segmentation_model_with_no_predictions`)?

lite/tests/segmentation/test_stability.py
```diff
@@ -0,0 +1,85 @@
+from random import choice
```
ntlind, line 1: we can no longer run all `lite` tests at once (using `pytest lite/tests/*`) since these test files share the same names as the test files in `lite/tests/classification/*`. can you update these file names to be unique?
image

lite/tests/segmentation/test_accuracy.py
```diff
@@ -0,0 +1,64 @@
...
+    loader.add_data(segmentations_from_boxes)
+    evaluator = loader.finalize()
+
+    metrics = evaluator.evaluate(as_dict=True)
+
+    actual_metrics = [m for m in metrics[MetricType.Accuracy]]
+    expected_metrics = [
+        {
+            "type": "Accuracy",
+            "value": 0.3334,
```
ntlind, line 43: can you help me understand the math behind .3334? I would have thought the accuracy would be higher given there's so many true negatives in these 270k pixel images.

lite/valor_lite/segmentation/computation.py
```diff
@@ -0,0 +1,179 @@
+import numpy as np
+from numpy.typing import NDArray
+
+
+def compute_intermediates(
```
ntlind, line 5: update this name to be more specific (e.g., `compute_intermediate_confusion_matrix`)

lite/tests/segmentation/test_stability.py
```diff
@@ -0,0 +1,85 @@
...
+
+import numpy as np
+from valor_lite.segmentation import Bitmask, DataLoader, Segmentation
+
+
+def generate_random_segmentations(
```
ntlind, line 7: nit: make private with an underscore before the name?

lite/tests/segmentation/test_precision.py
```diff
@@ -0,0 +1,96 @@
...
+    loader = DataLoader()
+    loader.add_data(segmentations_from_boxes)
+    evaluator = loader.finalize()
+
+    metrics = evaluator.evaluate(as_dict=True)
+
+    actual_metrics = [m for m in metrics[MetricType.F1]]
+    expected_metrics = [
+        {
+            "type": "F1",
```
ntlind, line 49: should this test precision instead of F1?

lite/tests/segmentation/test_dataloader.py
czaloom: This isn't possible since the dataclass validator catches a shape mismatch between groundtruths and predictions.

czaloom: See `test_annotation.py::test_segmentation`

lite/tests/segmentation/test_accuracy.py
czaloom: Updated the computation to account for background pixels being counted as True-Negatives (TN)

lite/tests/segmentation/test_dataloader.py
ntlind: passing an empty set of predictions was possible in the API, so I see this mismatch causing issues when you go to swap the API's compute functions with `lite`. can you check with Eric to confirm he's cool with no longer allowing empty prediction sets? if yes, then lgtm.

lite/tests/segmentation/test_stability.py
ntlind: as a user, having to run all of the tests separately before committing is a major pain. you may be able to [find a workaround here](https://docs.pytest.org/en/stable/explanation/goodpractices.html#choosing-an-import-mode) if you want to keep the names the exact same, but I don't think this is a pytest best practice.

> This allows the tracking of coverage to be separate

you'll still be able to track coverage separately since you're testing individual directories

> finding equivalent tests in another task type easier

you can append "_clf", "_det", "_seg" to the file names if you want. I don't see this benefit outweighing the annoyance of having to run all of the tests separately.

czaloom: Adding the init files got `pytest lite/tests` working again. I had removed these b/c I thought they werent necessary.

lite/tests/segmentation/test_dataloader.py
ntlind: thanks for checking!

ntlind (approved):

PR #775, opened 2024-10-02
Remove label keys from `lite` detections

ntlind:
## Improvements
- Removed label keys from all functions and tests
- Added a DeprecationWarning to the `...from_valor_dict` inner function
- Turn off some API benchmarks since a) they are failing our GH workflows and b) they aren't offering us useful information given the upcoming API changes

### DIFF
--- a/integration_tests/benchmarks/object-detection/benchmark_script.py
+++ b/integration_tests/benchmarks/object-detection/benchmark_script.py
@@ -301,8 +301,8 @@ def run_benchmarking_analysis(
     chunk_size: int = -1,
     ingestion_timeout: int = 30,
     evaluation_timeout: int = 30,
-    compute_pr: bool = True,
-    compute_detailed: bool = True,
+    compute_pr: bool = False,
+    compute_detailed: bool = False,
 ):
     """Time various function calls and export the results."""
     current_directory = Path(__file__).parent
--- a/integration_tests/benchmarks/object-detection/benchmark_script.py
+++ b/integration_tests/benchmarks/object-detection/benchmark_script.py
@@ -245,48 +245,52 @@ class Benchmark:
                     "model": f"{round(self.pd_deletion, 2)} seconds",
                 },
             },
-            "base+pr": {
-                "ingestion": {
-                    "dataset": f"{round(self.gt_ingest, 2)} seconds",
-                    "model": f"{round(self.pd_ingest, 2)} seconds",
-                },
-                "finalization": {
-                    "dataset": f"{round(self.gt_finalization, 2)} seconds",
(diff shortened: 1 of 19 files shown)


integration_tests/benchmarks/object-detection/benchmark_script.py
czaloom, line 1: Could you instead comment this out? I think we will need it once we merge lite into the API

lite/valor_lite/detection/computation.py
```diff
@@ -272,9 +276,9 @@ def compute_metrics(
 
     Returns
     -------
-    tuple[NDArray, NDArray, NDArray NDArray]
+    tuple[NDArray, NDArray, NDArray float]
```
czaloom, line 279: typo: missing comma

lite/valor_lite/detection/computation.py
```diff
@@ -272,9 +276,9 @@ def compute_metrics(
...
         Average Precision results.
-    tuple[NDArray, NDArray, NDArray NDArray]
+    tuple[NDArray, NDArray, NDArray float]
```
czaloom, line 281: typo: missing comma

integration_tests/benchmarks/object-detection/benchmark_script.py
ntlind: sure, the only change to this file is to set the default for compute_pr, etc. to False

lite/tests/detection/test_stability.py
```diff
@@ -17,7 +17,7 @@ def bbox(is_prediction):
             xmax,
             ymin,
             ymax,
-            [("class", choice(labels)), ("category", choice(labels))],
+            [choice(labels), choice(labels)],
```
czaloom, line 20: This doesnt make sense anymore, there should only be one groundtruth per box

lite/tests/detection/test_filtering.py
czaloom, line 1: With the modification to the basic detections fixture the label filter tests in this file aren't tested enough. Could you make a fixture with multiple labels and see that the filters perform the same?

lite/tests/detection/test_average_precision.py
```diff
@@ -37,60 +37,52 @@ def test__compute_average_precision():
...
             [1.0],  # iou = 0.1
             [1 / 3],  # iou = 0.6
         ]
     )
-    assert expected_ap.shape == average_precision.shape
-    assert np.isclose(average_precision, expected_ap).all()
+    assert expected.shape == average_precision.shape
+    assert np.isclose(average_precision, expected).all()
 
-    # since only one class, ap == map
```
czaloom, line 49: this comment might be good to retain

czaloom (approved):

PR #776, opened 2024-10-04
Remove label keys from `lite` classification

ntlind:
## Improvements
- Update `lite` tests and functions to no longer use label keys

### DIFF
--- a/lite/tests/classification/conftest.py
+++ b/lite/tests/classification/conftest.py
@@ -322,7 +322,7 @@ def classifications_multiclass() -> list[Classification]:
 
 
 @pytest.fixture
-def classifications_multiclass_true_negatives_check_animals() -> (
+def classifications_multiclass_true_negatives_check() -> (
     list[Classification]
 ):
     return [
@@ -330,20 +330,6 @@ def classifications_multiclass_true_negatives_check_animals() -> (
             uid="uid1",
             groundtruths=["ant"],
             predictions=["ant", "bee", "cat"],
-            scores=[0.15, 0.48, 0.37],
-        ),
-    ]
-
-
-@pytest.fixture
-def classifications_multiclass_true_negatives_check_ingredients() -> (
-    list[Classification]
-):
-    return [
-        Classification(
-            uid="uid2",
-            groundtruths=["egg"],
-            predictions=["egg", "milk", "flour"],
             scores=[0.15, 0.48, 0.37],
         ),
     ]
--- a/lite/tests/classification/conftest.py
+++ b/lite/tests/classification/conftest.py
@@ -7,34 +7,34 @@ def basic_classifications() -> list[Classification]:
     return [
         Classification(
             uid="uid0",
-            groundtruths=[("class", "0")],
+            groundtruths=["0"],
             predictions=[
-                ("class", "0"),
-                ("class", "1"),
-                ("class", "2"),
-                ("class", "3"),
+                "0",
+                "1",
+                "2",
+                "3",
             ],
             scores=[1.0, 0.0, 0.0, 0.0],
         ),
         Classification(
             uid="uid1",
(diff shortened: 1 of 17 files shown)


lite/tests/classification/conftest.py
```diff
@@ -173,35 +171,24 @@ def classifications_image_example() -> list[Classification]:
         Classification(
             uid="uid5",
             groundtruths=[
-                ("k4", "v4"),
-                ("k5", "v5"),
```
ntlind, line 177: I cut the "k5" and "k3" entries out of this data, rather than replicating our tests 3x everywhere this data is used, because these keys didn't seem worth repeating as separate, unique tests: they both just test for a false negative / false positive combo, which is already tested in "uid5" by gt=['v4'] and pd=['v1', 'v8']

lite/tests/classification/test_accuracy.py
```diff
@@ -197,61 +211,33 @@ def test_accuracy_with_example(
...
-            "parameters": {
-                "score_thresholds": [0.0],
-                "hardmax": True,
-                "label": {"key": "k5", "value": "v5"},
-            },
-        },
-        {
-            "type": "Accuracy",
-            "value": [0.0],
+            "value": [0.5],  # TODO add comment about the bug here
```
ntlind, line 236: this value used to be .333, which (I believe) was a bug: if you look at `test_evaluate_image_clf` in `integration_tests/...`, this value should have been .5. no action needed, just explaining why this changed

image

lite/valor_lite/classification/computation.py
```diff
@@ -70,14 +59,7 @@ def _compute_rocauc(
...
     # compute mean rocauc
-    summed_rocauc = np.bincount(label_metadata[:, 2], weights=rocauc)
-    mean_rocauc = np.zeros(n_label_keys, dtype=np.float64)
-    np.divide(
-        summed_rocauc,
-        count_labels_per_key,
-        where=count_labels_per_key > 1e-9,
-        out=mean_rocauc,
-    )
+    mean_rocauc = sum(rocauc) / n_labels
```
czaloom, line 62: Could you change this to `rocauc.mean()` to take advantage of the numpy representation?

lite/valor_lite/classification/manager.py
```diff
@@ -580,77 +534,50 @@ def _add_datum(self, uid: str) -> int:
...
-            self._evaluator.label_key_to_index[label[0]],
-        )
+        return self._evaluator.label_to_index[label]
 
     def _add_data(
         self,
         uid_index: int,
-        keyed_groundtruths: dict[int, int],
-        keyed_predictions: dict[int, list[tuple[int, float]]],
+        groundtruths: int,
```
czaloom, line 563: nit: rename to groundtruth since it is now singular

lite/tests/classification/conftest.py
```diff
@@ -337,42 +322,40 @@ def classifications_multiclass() -> list[Classification]:
...
+            groundtruths=["ant"],
+            predictions=["ant", "bee", "cat"],
             scores=[0.15, 0.48, 0.37],
         ),
     ]
 
 
 @pytest.fixture
-def classifications_multiclass_zero_count() -> list[Classification]:
+def classifications_multiclass_true_negatives_check_ingredients() -> (
```
czaloom, line 339: We can remove this fixture and its tests b/c it contains the same info as `classifications_multiclass_true_negatives_check_animals`

lite/tests/classification/test_accuracy.py
czaloom: Great catch! The 1/3 accuracy is b/c we are dividing by `n_datums` in the computation. This resulted in a flawed output since that label key didnt exist for one datum.

czaloom (approved):

PR #777, opened 2024-10-07
Enforce Single Groundtruth in Lite

czaloom:
Since we've removed label keys in detection and classification there should never be a condition where multiple groundtruths exist per annotation. 

# Changes
- Enforce a single label on ground truth annotations for detection and classification.

### DIFF
--- a/lite/valor_lite/detection/annotation.py
+++ b/lite/valor_lite/detection/annotation.py
@@ -17,7 +17,7 @@ class BoundingBox:
     def __post_init__(self):
         if len(self.scores) == 0 and len(self.labels) != 1:
             raise ValueError(
-                "If no scores are defined, then this is a ground truth and a single label requirement is enforced."
+                "Ground truths must be defined with no scores and a single label. If you meant to define a prediction, then please include one score for every label provided."
             )
         if len(self.scores) > 0 and len(self.labels) != len(self.scores):
             raise ValueError(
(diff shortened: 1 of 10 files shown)


lite/valor_lite/detection/annotation.py
```diff
@@ -15,6 +15,10 @@ class BoundingBox:
     scores: list[float] = field(default_factory=list)
 
     def __post_init__(self):
+        if len(self.scores) == 0 and len(self.labels) != 1:
+            raise ValueError(
+                "If no scores are defined, then this is a ground truth and a single label requirement is enforced."
```
ntlind, line 20: can you change this error message to be more explicit? something like "Ground truths must be defined with no scores and a single label. If you meant to define a prediction, then please include one score for every label provided."

lite/valor_lite/detection/annotation.py
```diff
@@ -34,6 +38,11 @@ class Polygon:
     def __post_init__(self):
         if not isinstance(self.shape, ShapelyPolygon):
             raise TypeError("shape must be of type shapely.geometry.Polygon.")
+
+        if len(self.scores) == 0 and len(self.labels) != 1:
```
ntlind, line 42: make more explicit; see last comment

lite/valor_lite/detection/annotation.py
```diff
@@ -34,6 +38,11 @@ class Polygon:
...
+            raise ValueError(
+                "If no scores are defined, then this is a ground truth and a single label requirement is enforced."
```
ntlind, line 44: make more explicit; see last comment

lite/valor_lite/detection/annotation.py
```diff
@@ -63,6 +72,19 @@ class Bitmask:
...
+
+        if (
+            not isinstance(self.mask, np.ndarray)
+            or self.mask.dtype != np.bool_
+        ):
+            raise ValueError(
+                "Expected mask to be of type `NDArray[np.bool_]`."
+            )
+
+        if len(self.scores) == 0 and len(self.labels) != 1:
```
ntlind, line 84: make more explicit; see last comment

ntlind (approved):

PR #778, opened 2024-10-08
Performance Bugfixes

czaloom:
# Issues
- Runtime for Object Detection doubled due to a data duplication bug in #775
- Unhandled `RuntimeWarning` in mAR and mAP computation due to removal of keys.

# Improvements
- All
  - Replaced `np.floating` with `np.float64` since the former sometimes raises warnings.
- Object Detection
  - Polygon IoU computation is now vectorized.
  - Bitmask IoU computation is now vectorized.

# Fixes
- Object Detection
  - Fixed data duplication bug
  - Fixed bitmask IoU testing where bitmasks with dtype `object` were being used in place of `np.bool_`
  - Fixed typing error. `type[...]` should be `Type[...]` to not throw linting errors.


lite/valor_lite/detection/computation.py
```diff
@@ -86,23 +89,41 @@ def compute_bitmask_iou(data: NDArray[np.floating]) -> NDArray[np.floating]:
...
+    np.divide(
+        intersection_,
+        union_,
+        where=union_ >= 1e-9,
+        out=ious,
+    )
+    return ious
 
 
 def compute_polygon_iou(
```
ntlind, line 124: nit: rename these functions to be plural?

ntlind (approved):

PR #783, opened 2024-10-09
Prepare Lite for Core Migration

czaloom:
# Changes
- Renamed `detection` to `object_detection`
- Renamed `segmentation` to `semantic_segmentation`
- Added stubs for future llm metrics.


lite/README.md
```diff
@@ -1 +1,141 @@
-# valor-lite: Compute classification, object detection, and segmentation metrics locally.
+# valor-lite: Fast, local machine learning evaluation.
+
+Valor-lite is a lightweight, numpy-based library designed for fast and seamless evaluation of machine learning models. It is optimized for environments where quick, responsive evaluations are essential, whether as part of a larger service or embedded within user-facing tools. By operating without the latency of external systems, Valor-lite ensures an efficient, smooth experience for users.
```
ntlind, line 3: > By operating without the latency of external systems, Valor-lite ensures an efficient, smooth experience for users.

suggest deleting this line. different external systems will have different "latencies", so it's not clear which specific system you're talking about here. this line also seems to convey the same information as the line above.

lite/README.md
```diff
@@ -1 +1,141 @@
-# valor-lite: Compute classification, object detection, and segmentation metrics locally.
+# valor-lite: Fast, local machine learning evaluation.
```
ntlind, line 1: suggest leaving valor-lite uncapitalized everywhere (i.e., no "Valor-lite")

lite/README.md
```diff
@@ -1 +1,141 @@
...
+
+Valor-lite is a lightweight, numpy-based library designed for fast and seamless evaluation of machine learning models. It is optimized for environments where quick, responsive evaluations are essential, whether as part of a larger service or embedded within user-facing tools. By operating without the latency of external systems, Valor-lite ensures an efficient, smooth experience for users.
+
+Valor, the broader platform, acts as a centralized evaluation store that simplifies the measurement, exploration, and ranking of model performance. This powerful tool helps data scientists and engineers continuously assess and refine their machine learning pipelines, enabling them to make data-driven decisions for ongoing improvement.
```
ntlind, line 5: can you add a note that the Valor service uses valor-lite to do its computations, and that we published valor-lite separately in case users want to evaluate performance without managing that separate service?

nit: some of these other benefits could live in the repo README.md, rather than in `lite/README.md`; talking about them both here seems confusing.

lite/pyproject.toml
```diff
@@ -12,6 +12,9 @@ dependencies = [
     "requests",
     "numpy",
     "shapely",
+    "nltk",
```
ntlind, line 15: can we make nltk and rouge_score optional as well?

lite/benchmarks/benchmark_objdet.py
```diff
@@ -8,7 +8,7 @@
 
 import requests
 from tqdm import tqdm
-from valor_lite.detection import DataLoader, MetricType
+from valor_lite.cv.detection import DataLoader, MetricType
```
ntlind, line 11: the "CV" bucket doesn't feel MECE: if the user does `pip install valor-<name>[cv]` (from your example yesterday), they'd expect classification to be installed too. why not leave classification, detection, and segmentation as-is (i.e., not grouping them under CV), but group the LLM metrics separately under NLP?

czaloom: Polling for feedback on flattening the import structure! 

As for the custom installs, I feel thats something we can deal with at a later point.

lite/pyproject.toml
czaloom: Im just going to omit these for now seeing that the llm metrics aren't added yet

lite/README.md
czaloom: trimmed it!

ntlind (approved):

PR #788, opened 2024-10-09
Remove `metrics_to_return`

czaloom:
# Issue
- `metrics_to_return` parameter obfuscates which metrics are computed together.

# Changes
- Remove option to specify `metrics_to_return`
- `evaluate` now returns **all** available metrics.
- Object Detection
  - Add `compute_precision_recall`
  - Change `_compute_confusion_matrix` to `compute_confusion_matrix`
- Classification
  - Add `compute_precision_recall`
  - Change `_compute_confusion_matrix` to `compute_confusion_matrix`
- Semantic Segmentation
  - Add `compute_precision_recall_iou`


ntlind:
I disagree with replacing the '.evaluate()` method for the following reasons:
- Calling multiple `compute...` methods instead of just one `.evaluate()` method is a worse user experience. As we add new metrics (which is somewhat likely on the LLM side), the number of independent calls that a user has to make will increase.
    - Ben's LLM metrics already involve 12 independent computations that would require separate `compute...` calls under this new pattern. We've heard from George that having to make separate calls for each LLM metric would constitute a poor user experience.
- As a user, I care about what specifically gets returned, not how the metrics are computed under the hood. Grouping multiple metrics under `compute_metrics` is more vague than allowing me to be specific about exactly what I want.  
- This change is purely preferential given `ConfusionMatrix` is already not included on the default metric list for `metrics_to_return`.  Making this change isn't necessary to split `ConfusionMatrix` into a separate calculation, and won't lead to significant performance benefits over just passing `ConfusionMatrix` into `metrics_to_return`.

Happy to be overruled if you and Eric agree!

czaloom:
> I disagree with replacing the '.evaluate()` method for the following reasons:
> 
> * Calling multiple `compute...` methods instead of just one `.evaluate()` method is a worse user experience. As we add new metrics (which is somewhat likely on the LLM side), the number of independent calls that a user has to make will increase.
>   
>   * Ben's LLM metrics already involve 12 independent computations that would require separate `compute...` calls under this new pattern. We've heard from George that having to make separate calls for each LLM metric would constitute a poor user experience.
> * As a user, I care about what specifically gets returned, not how the metrics are computed under the hood. Grouping multiple metrics under `compute_metrics` is more vague than allowing me to be specific about exactly what I want.
> * This change is purely preferential given `ConfusionMatrix` is already not included on the default metric list for `metrics_to_return`.  Making this change isn't necessary to split `ConfusionMatrix` into a separate calculation, and won't lead to significant performance benefits over just passing `ConfusionMatrix` into `metrics_to_return`.
> 
> Happy to be overruled if you and Eric agree!

After discussing this with Sean and Eric I would like to continue with the removal of metrics_to_return.

The pattern we will now be following for direct computation is:
```
evaluator.compute_precision_recall(...) # base metrics
evaluator.compute_confusion_matrix(...) # detailed metrics
```
 
The `evaluate` method will still exist without `metrics_to_return`. It will act as the user-facing function when used in notebooks. The default behavior of `evaluate` going forward will be to return all available metrics to the user.

ntlind:
I like this idea better than deleting `.evaluate()`, but this still doesn’t help with Ben’s LLM use case (where a user will very likely want to return a subset of LLM metrics at once, but won’t want to run 6-10 separate functions to get the metrics they need)

consider leaving `metrics_to_return` so that we `.evaluate()` will take common parameters across each task type, but making it so the user can also call `evaluator.compute_precision_recall`, `evaluator.compute_confusion_matrix` if they want something specific

ekorman (approved):

PR #789, opened 2024-10-09
Error handling for text gen in Valor

bnativi:
Error handling in valor for text generation:
- a new attribute "status" is added for all text generation parameters, which can be either "success" or "error"
- if an invalid LLM response error is thrown, then instead of crashing the whole evaluation, instead the metric is saved with status "error"
- if the status is "error", then the metric value will be None
- validation updated to work with these statuses
- all valor streaming manager joint_df values to be null in the event that a metric computation failed
- testing added for the statuses

Bug fix for LLM retries
- llm_api_params previously only allowed strings or dictionaries as values, but retries is an integer parameter. The validation was updated to allow retries to be an integer


core/valor_core/metrics.py
```diff
@@ -668,18 +668,34 @@ class _TextGenerationMetricBase:
...
 
-    value: int | float | dict[str, float]
+    status: str
+    value: int | float | dict[str, float] | None
     parameters: dict
     __type__ = "BaseClass"
 
     def __post_init__(self):
         """Validate instantiated class."""
+        if not isinstance(self.status, str):
```
czaloom, line 686: I would maybe prefer the use of an enum class here instead of a raw string.

ntlind (approved):

czaloom (approved):

PR #798, opened 2024-10-15
Lite Classification Accuracy Fix

czaloom:
# Changes
- The `Accuracy` metric is now computed over all labels rather than being per-label.

### DIFF
--- a/.github/workflows/client-api-tests-and-coverage.yml
+++ b/.github/workflows/client-api-tests-and-coverage.yml
@@ -134,7 +134,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: ["3.8", "3.9", "3.10", "3.11"]
+        python: ["3.7.1", "3.8", "3.9", "3.10", "3.11"]
     defaults:
       run:
         working-directory: ./client
--- a/.github/workflows/client-api-tests-and-coverage.yml
+++ b/.github/workflows/client-api-tests-and-coverage.yml
@@ -134,7 +134,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python: ["3.8", "3.9", "3.10", "3.11"]
     defaults:
       run:
         working-directory: ./client
--- a/.github/workflows/client-api-tests-and-coverage.yml
+++ b/.github/workflows/client-api-tests-and-coverage.yml
@@ -131,10 +131,10 @@ jobs:
   client-unit-tests:
     env:
       COVERAGE_FILE: .coverage.client-unit-tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python: ["3.7.1", "3.8", "3.9", "3.10", "3.11"]
+        python: ["3.7", "3.8", "3.9", "3.10", "3.11"]
     defaults:
       run:
         working-directory: ./client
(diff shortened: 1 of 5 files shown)


.github/workflows/client-api-tests-and-coverage.yml
```diff
@@ -131,7 +131,7 @@ jobs:
   client-unit-tests:
     env:
       COVERAGE_FILE: .coverage.client-unit-tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
jyono, line 134: where are your runners hosted? just using github ones? why wouldn't you use `24.04`?

lite/tests/classification/test_accuracy.py
```diff
@@ -53,18 +53,12 @@ def test_accuracy_computation():
     )
 
     # score threshold, label, count metric
-    assert accuracy.shape == (2, 4)
+    assert accuracy.shape == (2,)
```
jyono, line 56:
```suggestion
    assert accuracy.shape == (2)
```
delete the trailing comma

lite/tests/classification/test_accuracy.py
```diff
@@ -53,18 +53,12 @@ def test_accuracy_computation():
...
 
     # score >= 0.25
-    assert accuracy[0][0] == 2 / 3
-    assert accuracy[0][1] == 1.0
-    assert accuracy[0][2] == 2 / 3
-    assert accuracy[0][3] == 1.0
+    assert accuracy[0] == 2 / 3
```
jyono, line 59: you should use 
`math.isclose()`

jyono (commented):
looks mostly good to me.

jyono (commented):
https://github.com/Striveworks/valor/pull/798/files#diff-51c917d730bcc12aa79cec4113bd3324722e7be8271807b18d151b524a4ec70aR124

this is a very long function testing it requires testing a whole lot of things. 

It might make sense to pull out the different calculations into their own functions and test this rigorously in isolation.

.github/workflows/client-api-tests-and-coverage.yml
czaloom: We are using the github ones. It was previously pinned to `ubuntu-latest` which I am assuming was `22.04` until this weekend. 

The reason for this chnage is that `python3.7` is not a valid `24.04` target.

lite/tests/classification/test_accuracy.py
czaloom: The comma is actually correct here 😅 

`(2)` evaluates to `2` of type `int`
`(2,)` evaluates to `(2,)` of type `tuple[int]`

lite/valor_lite/classification/computation.py
```diff
@@ -182,9 +182,9 @@ def compute_metrics(
         out=precision,
     )
 
-    accuracy = np.zeros_like(recall)
+    accuracy = np.zeros(n_scores, dtype=np.float64)
```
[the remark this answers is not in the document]
czaloom: > because we aren't considering labels
This is why haha

`recall` has shape `(n_scores, n_labels)` so we need to explicitly define size without a label count.

lite/tests/classification/test_accuracy.py
czaloom: I understand that this would be best practive.

However, this will require changing a lot of testing since metrics are dense dictionaries and it has not been a problem up to this point.

.github/workflows/client-api-tests-and-coverage.yml
jyono: why do we care? python 3.7 is EOL

czaloom: Im good with removing it, the valor-client code is not relevant to this pr so i just forced that version.

rsbowman-striveworks (approved):

PR #800, opened 2024-10-15
Remove Old Python Versions and Update Client

jyono:
Removing EOL python versions and checking support for modern LTS ones
https://devguide.python.org/versions/


czaloom (approved):

czaloom:
Updated repo settings to remove python 3.8 requirement. 

The testing requirement is now set at python 3.10

PR #817, opened 2024-11-08
Bugfix - Object Detection IOU

czaloom:
### valor version checks

- [X] I have confirmed this bug exists on the latest version of valor.

### Reproducible Example

```python
def test_iou_computation():

    detection = Detection(
        uid="uid",
        groundtruths=[
            BoundingBox(xmin=0, xmax=10, ymin=0, ymax=10, labels=["0"]),
            BoundingBox(xmin=100, xmax=110, ymin=100, ymax=110, labels=["0"]),
            BoundingBox(
                xmin=1000, xmax=1100, ymin=1000, ymax=1100, labels=["0"]
            ),
        ],
        predictions=[
            BoundingBox(
                xmin=1,
                xmax=11,
                ymin=1,
                ymax=11,
                labels=["0", "1", "2"],
                scores=[0.5, 0.25, 0.25],
            ),
            BoundingBox(
                xmin=105,
                xmax=116,
                ymin=105,
                ymax=116,
                labels=["0", "1", "2"],
                scores=[0.5, 0.25, 0.25],
            ),
        ],
    )

    loader = DataLoader()
    loader.add_bounding_boxes([detection])

    assert len(loader.pairs) == 1

    # show that three unique IOUs exist
    unique_ious = np.unique(loader.pairs[0][:, 3])
    assert np.isclose(
        unique_ious, np.array([0.0, 0.12755102, 0.68067227])
    ).all()
```

### Issue Description

There are two serious bugs.
1. IOU is being computed **after** label permutation which means that a ton of work is being redone.
2. (1) was introduced at some point without updating the IOU allocation. This means that all IOUs were identical for every pair within an image. This did not affect the results between images.

### Expected Behavior

IOU should be computed the minimum number of times and should be properly reported per groundtruth-prediction pair.

### DIFF
--- a/lite/valor_lite/object_detection/manager.py
+++ b/lite/valor_lite/object_detection/manager.py
@@ -1,17 +1,10 @@
 from collections import defaultdict
 from dataclasses import dataclass
-from typing import Type
 
 import numpy as np
-import valor_lite.object_detection.annotation as annotation
 from numpy.typing import NDArray
 from tqdm import tqdm
-from valor_lite.object_detection.annotation import (
-    Bitmask,
-    BoundingBox,
-    Detection,
-    Polygon,
-)
+from valor_lite.object_detection.annotation import Detection
 from valor_lite.object_detection.computation import (
     compute_bbox_iou,
     compute_bitmask_iou,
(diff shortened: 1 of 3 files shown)


lite/valor_lite/object_detection/manager.py
```diff
@@ -459,8 +428,8 @@ def _compute_ious_and_cache_pairs(
                         float(score),
                     ]
                 )
-                for pidx, plabel, score, _ in predictions
-                for gidx, glabel, _ in groundtruths
+                for pidx, plabel, score in predictions
+                for gidx, glabel in groundtruths
                 if ious[pidx, gidx] >= 1e-9
             ]
         )
```
rsbowman-striveworks, line 435: Here don't you just want:
```
pairs = [
    np.array(
        ...
    )
    for pidx, plabel, score in predictions
    # etc.
]
```
?

rsbowman-striveworks (approved):
:+1:

PR #847, opened 2025-05-30
Resolve OOM issue + better examples for classification

czaloom:
# Changes
- Resolves OOM issue
  - Reduced Confusion Matrix memory complexity from `(n_score_thresholds * n_labels * n_datums * n_examples)` to `(n_score_thresholds * n_labels * n_datums)`
- Remove `number_of_examples` option from classification
- Renamed `Exceptions` to `Errors`

### DIFF
--- a/src/valor_lite/classification/utilities.py
+++ b/src/valor_lite/classification/utilities.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 import numpy as np
 from numpy.typing import NDArray
 
+from valor_lite.classification.computation import PairClassification
 from valor_lite.classification.metric import Metric, MetricType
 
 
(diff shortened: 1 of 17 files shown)


src/valor_lite/classification/utilities.py
```diff
@@ -101,129 +102,110 @@ def unpack_precision_recall_rocauc_into_metric_lists(
...
-            "count": max(
-                int(unmatched_ground_truths[gt_label_idx, 0]),
-                0,
-            ),
-            "examples": [
+    for idx in range(n_max):
+        if idx < n_matched:
+            glabel = index_to_label[unique_matches[idx, 1]]
+            plabel = index_to_label[unique_matches[idx, 2]]
+            confusion_matrix[glabel][plabel]["count"] += 1
```
rsbowman-striveworks, line 155: now that we're giving all examples is it always true that the `count` here is just the length of the `examples` entry?  I worry about redundant stuff like this because if these get out of sync it's confusing...  Can we just return the `examples` list and let callers take the `len` of that?  Or is that a breaking interface change?

sasbury: I will let Charles answer but i think we are going to try to separate "here is the matrix" from "here are the examples" to allow the UI to be more svelt showing the first, and more precise on the second

czaloom: It is a bit redundant now but the intent is to eventually split the examples from the counts entirely. We are still implementing a number of examples cap on the chariot-valor side.

rsbowman-striveworks (approved):

PR #849, opened 2025-07-03
handle overlapped segmentations more gracefully

czaloom:
Instead of raising an error, segmentations with overlapped masks now operate on a first-come-first-serve basis in which the first label to arrive "claims" a pixel. The total number of affected pixels is counted and then conveyed in a warning after all labels have been checked.

### DIFF
--- a/src/valor_lite/semantic_segmentation/annotation.py
+++ b/src/valor_lite/semantic_segmentation/annotation.py
@@ -1,3 +1,4 @@
+import warnings
 from dataclasses import dataclass, field
 
 import numpy as np
@@ -80,7 +81,8 @@ class Segmentation:
         self.size = self.shape[0] * self.shape[1]
 
         mask_accumulation = None
-        for groundtruth in self.groundtruths:
+        mask_overlap_accumulation = None
+        for idx, groundtruth in enumerate(self.groundtruths):
             if self.shape != groundtruth.mask.shape:
                 raise ValueError(
                     f"ground truth masks for datum '{self.uid}' should have shape '{self.shape}'. Received mask with shape '{groundtruth.mask.shape}'"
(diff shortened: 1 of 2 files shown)


src/valor_lite/semantic_segmentation/annotation.py
```diff
@@ -80,29 +81,65 @@ def __post_init__(self):
...
             if self.shape != prediction.mask.shape:
                 raise ValueError(
-                    f"prediction masks for datum '{self.uid}' should have shape '{self.shape}'. Received mask with shape '{prediction.mask.shape}'"
+                    f"prediction masks for datum '{self.uid}' should have shape '{self.shape}'. Received mask with shape '{prediction.mask.shape}'",
+                    UserWarning,
                 )
 
             if mask_accumulation is None:
                 mask_accumulation = prediction.mask.copy()
+                mask_overlap_accumulation = np.zeros_like(mask_accumulation)
```
rsbowman-striveworks, line 125: Looks like the next chunk of code duplicates the previous chunk -- would it help to make this a function and call it twice?

src/valor_lite/semantic_segmentation/annotation.py
```diff
@@ -79,30 +80,39 @@ def __post_init__(self):
...
-                mask_accumulation = mask_accumulation | groundtruth.mask
+        self._validate_bitmasks("groundtruths")
+        self._validate_bitmasks("predictions")
 
+    def _validate_bitmasks(self, key: str):
         mask_accumulation = None
-        for prediction in self.predictions:
-            if self.shape != prediction.mask.shape:
+        mask_overlap_accumulation = None
+        for idx, annotation in enumerate(getattr(self, key)):
```
rsbowman-striveworks, line 89: Can we do it without the `getattr` call?  That makes it hard to do static analysis and seems like overkill here.  In this case we know exactly which lists we want to operate on so maybe we can just pass them as a parameter?


I think the `getattr` call makes it hard to do static analysis here -- if we change the name of "groundtruths" for example, we only see an error at runtime whereas if

